PR #7, opened 2025-10-26
GPU vs Moments LD statistics for the IM Model

ak1376:
Comparing the GPU LD statistics to the LD statistics from Moments in the IM model.


pg_gpu/ld_statistics.py
```diff
@@ -7,7 +7,7 @@
 
 import cupy as cp
 from typing import Optional, Union, Tuple, List, Dict
-
+dtype = cp.float64
```
andrewkern, line 10: so rather than do this and then in each function cast the counts array with `counts = cp.asarray(counts, dtype=dtype)` can't we just say `cp.set_default_dtype(cp.float64)` at this level? did you try that?

PR #10, opened 2026-01-16
Modifying LD stats to support three populations on the GPU

ak1376:
Adding to the Dz, Pi, and other functions to support three population LD statistics on the GPU. 
Some code may be duplicated, because I did not want to modify extensively the functions that worked for the two population case.

### DIFF
--- a/docs/source/examples.rst
+++ b/docs/source/examples.rst
@@ -65,9 +65,11 @@ Integration with moments
    h.sample_sets = {"pop1": list(range(10))}
    
    # Compute LD statistics with GPU acceleration
+   # Memory-efficient: only processes pairs within max distance
+   # chunk_size='auto' adapts to available GPU memory
    ld_stats = h.compute_ld_statistics_gpu_single_pop(
        bp_bins=[0, 1000, 5000],
-       pop="pop1"
+       chunk_size='auto'  # or int for fixed size (e.g., 500_000)
    )
    
    # Use with moments demographic models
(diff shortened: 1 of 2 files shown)


pg_gpu/ld_statistics.py
```diff
@@ -1,453 +1,316 @@
...
 """
 
+from __future__ import annotations
+
 import cupy as cp
-from typing import Optional, Union, Tuple, List, Dict
+from typing import Optional, Union, Tuple, List, Dict, Sequence
 
 
-def dd(counts: cp.ndarray,
```
andrewkern, line 12: why are you deleting this?

pg_gpu/ld_statistics.py
```diff
@@ -1,453 +1,316 @@
...
-            populations = (0, 0, 0)
-    
-    return _dz_multi(counts, populations, n_valid)
+    s = 4 * pop
+    inferred = cp.sum(counts[:, s:s + 4], axis=1)
 
+    if n_valid is None:
+        return inferred
 
-def pi2(counts: cp.ndarray,
```
andrewkern, line 99: again-- why are you deleting this?

andrewkern (changes requested):
this PR might work, but it's deleting a ton of well tested, working code. please revise your PR to only add the new three population functions, and to not delete the working two population code.

I'd recommend simply adding three population functions on top of what is already here.

PR #26, opened 2026-04-06
use GPU hash path for haplotype_count, add profiling scripts

kevinkorfmann:
## Summary                                                                                                         
  - Use GPU dot-product hashing for `haplotype_count()` instead of CPU string conversion                             
  - Reuses the same hash approach already used by `haplotype_diversity()` (two random projections + lexsort)         
  - Falls back to CPU clustering when missing data is present                                                        
  - ~300x speedup for typical dataset sizes                                                                          
                                                                                                                     
  | 200 | 50,000 | ~1,744 | 5.7 | ~300x |                                                                            
  | 100 | 50,000 | ~1,744 | 3.3 | ~530x |                                                                            
                                                                                                                     
  ## Test plan                                                                                                       
  - [ ] `pixi run pytest tests/test_diversity.py -v` (24 passed)                                                     
  - [ ] `pixi run python debug/bench_haplotype_count.py`

### DIFF
--- a/debug/bench_haplotype_count.py
+++ b/debug/bench_haplotype_count.py
@@ -1,0 +1,60 @@
+"""Benchmark haplotype_count: before (CPU string hashing) vs after (GPU dot-product hashing)."""
+
+import time
+import numpy as np
+import cupy as cp
+from pg_gpu import HaplotypeMatrix
+from pg_gpu.diversity import haplotype_count, haplotype_diversity
+
+
+def make_data(n_haps, n_snps, seed=42):
+    rng = np.random.default_rng(seed)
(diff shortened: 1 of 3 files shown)


andrewkern:
Thanks for getting this started! I picked up the idea and extended it in #89: that PR factors the hashing into a shared helper used by `haplotype_diversity`, `haplotype_count`, and `selection._distinct_haplotype_frequencies` (which had been a third line-for-line copy of the same algorithm). It also adds parity tests on real gamb data and exercises the CPU wildcard fallback with injected missingness.

Closing this in favor of #89.

PR #27, opened 2026-04-06
add Gram matrix fast path for pairwise_distance

kevinkorfmann:
## Summary                                                                                                         
  - Add Gram matrix fast path for `pairwise_distance()` with euclidean/sqeuclidean metrics                           
  - Uses `d²(i,j) = ||x_i||² + ||x_j||² - 2*x_i·x_j` via a single `X @ X.T` (n x n) instead of materializing all     
  (n_pairs, n_variants) difference vectors                                                                           
  - Falls back to batched pairwise approach for cityblock metric or when missing data is present                     
  - ~13x speedup, exact correctness (0.0 error vs scipy reference)                                                   
                                                                                                                     
  | 200 | 50,000 | ~75 | 5.8 | ~13x |                                                                                
  | 200 | 100,000 | ~150 | 17.1 | ~9x |                                                                              
                                                                                                                     
  ## Test plan                                                                                                       
  - [ ] `pixi run pytest tests/ -k 'decomposition or pca or pcoa or distance'` (26 passed)                           
  - [ ] `pixi run python debug/bench_pairwise_distance.py`

### DIFF
--- a/pg_gpu/decomposition.py
+++ b/pg_gpu/decomposition.py
@@ -368,26 +368,55 @@ def pairwise_distance(haplotype_matrix: HaplotypeMatrix,
         matrix.transfer_to_gpu()
 
     hap = matrix.haplotypes
+    from ._memutil import dac_and_n
+
+    _, n_valid = dac_and_n(hap)
 
     if missing_data == 'exclude':
-        missing_per_var = cp.sum(hap < 0, axis=0)
-        complete = missing_per_var == 0
+        complete = n_valid == hap.shape[0]
         hap = hap[:, complete]
-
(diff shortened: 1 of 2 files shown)


kevinkorfmann:
Rebased onto the latest main and simplified the implementation. Euclidean and squared-Euclidean distances now use the memory-safe chunked Gram path, while missing-data and cityblock cases retain the existing fallback.

nspope:
Sorry, this was rewritten to be multiallelic safe in #142 which is on the big multiallelic fix branch that's going to get merged in soon by #184. Can you wait until this is in, as it'll likely have merge conflicts?

PR #30, opened 2026-04-06
always use histogram kernel for iHS (2-8x faster than fused)

kevinkorfmann:
## Summary                                                                                                                           
  - Switch iHS to always use the histogram kernel instead of the fused kernel                                                          
  - The fused kernel (one block per variant, __syncthreads() per scan step) was slower than the histogram kernel (one thread per pair, 
  no synchronization) at all haplotype counts                                                                                          
  - 2.4-7.8x speedup depending on sample size, machine-precision accuracy                                                              
                                                                                                                                       
  ## Benchmarks (NVIDIA A100, 50k variants)                                                                                            
                                                                                                                                       
  | n_haps | Before (ms) | After (ms) | Speedup |                                                                                      
  |--------|-------------|------------|---------|                                                                                      
  | 100 | ~162 | 68.8 | 2.4x |                                                                                                         
  | 200 | ~598 | 76.6 | 7.8x |                                                                                                         
                                                                                                                                       
  ## Test plan                                                                                                                         
  - [ ] `pixi run pytest tests/test_selection.py tests/test_selection_allel_comparison.py -v` (27 passed)                              
  - [ ] `pixi run python debug/bench_ihs_kernels.py`


pg_gpu/selection.py
```diff
@@ -1403,7 +1403,7 @@ def _ihh01_scan_hist_gpu(h, gaps, min_ehh=0.05, min_maf=0.05,
     n_pairs = (n_haplotypes * (n_haplotypes - 1)) // 2
 
     if max_ssl_cap is None:
-        max_ssl_cap = min(n_variants, 10_000)
+        max_ssl_cap = min(n_variants, 2_000)
```
andrewkern: this isn't a safe ssl_cap on real data

andrewkern:
i benchmarked this on Ag1000G data and the ssl_cap change isn't ok. on Ag1000G reducing the ssl_cap to 2000 means the scan is only over ~8kb. That's not enough.

Some benchmarking against the ssl_cap=10000
  - ssl_cap=2000: corr=0.92, 19% of sites affected, max 84% relative error
  - ssl_cap=5000: corr=0.98, 12% of sites affected, max 58% relative error

the kernel switch might be okay however-- the issue with the histogram approach is memory. I'm looking in to that now

andrewkern:
the histogram with ssl_cap=10_000 completes without OOM on full Ag1000G 3R (2940 haplotypes, 10.9M variants). it's faster, but non-exact, which is why I wrote it to begin with. I think the thing to do here is to perhaps remove the commit that lowers the cap, but accept the switch.

Issue #76, opened 2026-04-25
Doc review


nspope:
as of commit 5322088

### Installation

- It'd be nice to have a "Hi there" section or pre-section text blurb, that says what the package is for and what it is inspired by (e.g. loosely modeled on scikit-allel but with GPU acceleration and more features)

- Requirements should include pixi and not mention cupy/numpy/etc. Or, there should be instructions on how to install without pixi. I'm under the impression we're conditioning on pixi install for the nvidia build tools. If we are conditioning on a pixi install it'd be good to say why ("we're not just being capricious, pixi makes the nvidia compilation toolchain portable and eliminates headaches from managing virtual environments!")

- There should be mention of how to run a script that uses pg_gpu here, e.g. `pixi run python script.py` -- this is shown for moments but should be in the first section. 

- I'm not sure why we're showing the moments integration here specifically---I suppose just to illustrate using the sub-environment with pixi? Why do we have a separate moments environment (do the additional dependencies really add that much overhead, or are they just needed for tests/examples)? If the latter, maybe we should title this section "Example-specific environments"

- More generally, make it clear to people who are coming from pip or pip-in-conda land how to use the package, e.g.:

  * What if I want to run a script, but not from the root of the pg_gpu repo?

  * What if I want to run pg_gpu from a notebook?

  * What if I want to use pg_gpu as a dependency in my package?

- "verify installation" should have the bash invocation to run the script (and put script name at top of code blurb, `# is_pg_gpu_installed.py`). Also why is this section needed? Could we simply check for cuda on init and error out (so import fails with informative message)?

### Quick start

- Generally: use longer variable names to give more of a lead here, e.g. "se" ==> "stderr"; "T, B" ==> "f3_numer, f3_denom"; and don't reuse names in the same blurb, e.g. "est" ==> "jack_est" and "boot_est"

- Loading data: Maybe say upfront that this is coming from diploid VCF and is interpreted as phased. Most folks are probably coming to this with diploid data, either phased or unphased.

- Patterson: why are numerator and denominator returned separately? This needs an intro sentence or comment to explain (and feeds into next section)

- Jackknife/bootstrap: "both estimators" ==> which estimators? estimators of what? needs an intro sentence or two, unlike the other stats it's not self-evident what's going on from the script blurb

- Bootstrap: where are 'num_blocks' and 'den_blocks' created? This workflow needs more hand-holding/context.

- Lostruct: why not wrap this workflow into a one-and-done like was done for Patterson's D?

[...]

PR #84, opened 2026-04-28
feat: side-by-side scikit-allel vs pg_gpu tutorial

andrewkern:
## Summary

New tutorial example that runs the same windowed-summary scan on
real Anopheles gambiae X-chromosome data using both scikit-allel
(three separate windowed calls) and pg_gpu (one fused
\`windowed_analysis\` call), verifies strict numerical agreement,
and reports the wall-clock speedup.

Closes the "scikit-allel side-by-side tutorial" comment from #76.

## What it adds

- \`examples/scikit_allel_comparison.py\` (~225 lines): main script
- \`docs/source/tutorials/scikit_allel_comparison.rst\`: tutorial
  page (added to the tutorials toctree after \`ld_blocks\`)
- \`tests/test_examples_run.py\`: smoke test that runs the script
  on the 4 Mb subset

## Three diversity stats

* \`allel.windowed_diversity\` ↔ \`pg_gpu\` \`'pi'\`
* \`allel.windowed_watterson_theta\` ↔ \`pg_gpu\` \`'theta_w'\`
* \`allel.windowed_tajima_d\` ↔ \`pg_gpu\` \`'tajimas_d'\`

(LD windowed scan was scoped out: scikit-allel's per-window pair-r²
is too slow at chromosome scale and the median-vs-mean / r²-vs-σ_D²
mismatch produced structurally weak agreement.)

## Verification

Strict numerical agreement (\`rtol=1e-5, atol=1e-8\`) on all three
stats. Trailing partial windows are masked because the two libraries
normalize them differently (allel divides per-site sums by the
actual span; pg_gpu divides by the fixed \`window_size\` parameter)
— the script flags it explicitly when this happens.

Observed agreement on the gamb full X chromosome (n=200
haplotypes, ~5.3 M variants, 2440 windows of 10 kb):

| stat | max abs diff |
|---|---|
| pi | 3.7e-17 |
| theta_w | 1.5e-16 |
| tajimas_d | 3.3e-14 |

## Timing

Wall-clock comparison on the full X (n_warmup=1, default 10 kb
windows, single A100, pixi-managed env):

| | scikit-allel | pg_gpu | speedup |
|---|---|---|---|
| time | 1.44s | 0.30s | 4.8x |

Modest because at 10 kb windows the compute is small enough that
allel's overhead dominates. Larger window sizes (more variants per
window) widen the gap.

## Figure

5-panel layout dropped to 4 panels with the LD scope reduction:
π / θ_W / Tajima's D traces overlaid (scikit-allel as a wider
lighter trace beneath pg_gpu's thinner darker one — agreement =
visual identity), plus a horizontal timing-bar panel at the bottom
annotated with the speedup ratio.

## Test plan

- [x] \`pixi run ruff check examples/ tests/\` clean
- [x] \`pixi run pytest tests/test_examples_run.py -v\` passes (~7s)
- [x] \`make -C docs html\` clean (no warnings)
- [x] \`pixi run python examples/scikit_allel_comparison.py --no-plot\`
      (full X) completes without OOM on a single A100 in <10s
- [x] \`pixi run python examples/scikit_allel_comparison.py -o /tmp/full.png\`
      (full X) renders the 4-panel figure

### DIFF
--- a/docs/source/tutorials.rst
+++ b/docs/source/tutorials.rst
@@ -19,6 +19,7 @@ time, see :doc:`examples` instead.
    tutorials/accessibility_mask
    tutorials/local_pca
    tutorials/ld_blocks
+   tutorials/scikit_allel_comparison
    tutorials/moments_integration
 
 Each tutorial follows the same structure:
--- a/docs/source/tutorials/scikit_allel_comparison.rst
+++ b/docs/source/tutorials/scikit_allel_comparison.rst
@@ -55,8 +55,19 @@ What the script does
    two libraries normalize partial windows slightly differently --
    a single-window cosmetic effect, not a real numerical
    disagreement).
-5. Subsamples ``--ld-snps`` (default 10,000) random SNPs and
-   computes a pairwise LD-decay curve on each side. The pg_gpu
+5. Selects a contiguous block of ``--ld-snps`` SNPs (default
+   10,000) centered on the chromosome midpoint, drops variants
+   below ``--ld-mac-min`` (default 10, i.e. MAF 0.05 in n=100
+   diploids), and computes a pairwise LD-decay curve on each side.
+   A contiguous block (vs. a random subsample) gives dense coverage
+   of short pair distances, which is where the LD-decay signal
+   lives -- a random 10k subsample leaves only a handful of pairs
+   in each short-distance bin and the curve washes out into noise.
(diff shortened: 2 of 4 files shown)


andrewkern:
Superseded by #87, which combined these commits with the fused-kernel work and squash-merged into main.

PR #96, opened 2026-05-11
including htslib because of bgzip use in some tests

mufernando:
I had a bunch of tests fail because of bgzip not being installed.


andrewkern:
i'll fix this

PR #97, opened 2026-05-11
Changes to pg-gpu tour

mufernando:
I ran and inspected the pg-gpu tour ipynb. For now I just changed:
- [x] Comparison of Zarr vs. VCF loading times

Outstanding stuff I need your input on:
- [ ] At the top there is a mention of a GPU acceleration section that does not exist; do you want to compare CPU vs. GPU compute times somehow here?
- ~~Should we include an accessibility mask example? NO, because this is explored in other parts of the docs.~~

### DIFF
--- a/examples/pg_gpu_tour.ipynb
+++ b/examples/pg_gpu_tour.ipynb
@@ -65,7 +65,17 @@
    "id": "nq0eyf8lw2",
    "metadata": {},
    "outputs": [],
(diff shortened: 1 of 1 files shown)


andrewkern:
is this ready for review?

mufernando:
Yes, but I don't know what to do about the alleged section 2 comparing GPU performance. What did you have in mind there? I can also remove the mention of this section altogether and re-number the sections.

andrewkern:
we can do away with it at this point

mufernando:
OK, this is ready for review.

PR #99, opened 2026-05-11
Revising the Examples

mufernando:
I carefully reviewed and ran all the Example scripts.

Here are my main changes:
- Fix a factual error in admixture_detection doc: the expected D under gene flow is negative, not positive. Also adds a sentence noting the leftward shift in the per-block D distribution due to genome-wide admix.
- Tighten wording in accessibility_mask: "exon" → "region" (the simulated block is a generic low-callability region, not specifically an exon
- Make explicit that the change is between a moments.LD workflow and a pg_gpu-accelerated one; fix "only line" → "only part" to match the actual multi-line loop.
- Add comments to examples/sweep_tajimas_d_bootstrap.py explaining why duplicate positions are shifted.

One thing that needs to be dealt with still:
- [ ] The full dataset hard-coded in the scikit-allel vs pg-gpu script does not exist (`gamb.X.phased.n100.zarr`)

### DIFF
--- a/docs/source/tutorials/accessibility_mask.rst
+++ b/docs/source/tutorials/accessibility_mask.rst
@@ -42,7 +42,7 @@ known, then verifies that the mask recovers it:
 
 1. Simulate a 1 Mb chromosome under ``msprime`` with a 200 kb central
    block of 100x lower mutation rate -- a stand-in for a low-callability
-   exon. The genealogy is identical inside and outside the block, so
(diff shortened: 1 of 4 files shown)


mufernando:
Note this is ready for review. The outstanding issue that I cannot fix is the missing dataset.

andrewkern (approved):
LGTM

PR #101, opened 2026-05-15
Transparent biobank-scale zarr streaming for HaplotypeMatrix and GenotypeMatrix

andrewkern:
## Summary

Adds transparent biobank-scale streaming behind `HaplotypeMatrix.from_zarr` and `GenotypeMatrix.from_zarr`. A store too big to fit eagerly on the GPU now returns a `StreamingHaplotypeMatrix` / `StreamingGenotypeMatrix` that walks the chromosome chunk-by-chunk through the same public APIs: `windowed_analysis`, `sfs.sfs`, `sfs.joint_sfs`, `compute_ld_statistics_gpu_single_pop`/`_two_pops`, `ibs`, `grm`. GPU peak memory scales with one chunk, not the chromosome.

Tested end-to-end on `stdpopsim` `OutOfAfrica_2T12` at chr15 / 50 k diploids per population (200 k haplotypes, 11.6 M variants, 7.9 GB bio2zarr/VCZ store): the full per-window + SFS + Garud + LD-decay + r² heatmap scan completes in about **16 minutes** on a single A100 80 GB.

## What's added

**Streaming infrastructure**
- `pg_gpu/zarr_source.py`: `ZarrGenotypeSource` opens a VCZ store and exposes `slice_region`, `slice_subsample`, `slice_region_gpu` (kvikio GPU-decode), `slice_subsample_gpu` (kvikio + GPU ploidy gather, ~100× faster than the host-side `oindex` at biobank-scale sample subsets).
- `pg_gpu/streaming_matrix.py`: `_StreamingMatrixBase` + `StreamingHaplotypeMatrix` + `StreamingGenotypeMatrix`, plus `ChunkFetcher` with `HostChunkFetcher` and `KvikioChunkFetcher` (zstd / blosc / lz4 / deflate decode on the GPU via nvCOMP).
- `pg_gpu/_gpu_genotype_prep.py`: `build_haplotype_matrix` and `build_genotype_matrix` do the ploidy interleave + transpose on the GPU.
- `pg_gpu/_biobank_warning.py`: `BiobankScaleWarning` on `from_vcf` for stores or sample counts above the biobank threshold.
- `pg_gpu/zarr_io.py`: `allel_zarr_to_vcz` streams a scikit-allel store to vcz layout in fixed-size variant blocks (does not materialize the full matrix), so empirical data in allel layout converts on modest hardware.

**Streaming-aware kernels** (existing call sites work unchanged on a streaming matrix)
- `windowed_analysis(stream, ...)` and `sfs.sfs` / `sfs.joint_sfs(stream, ...)`: per-chunk dispatch with sum reduction.
- `compute_ld_statistics_gpu_single_pop` and `_two_pops` on `StreamingHaplotypeMatrix`: tail-buffer stitch + (tail × tail) pair mask, so pairs that straddle a chunk boundary are captured exactly once.
- `ibs` and `grm` on streaming matrices: variant-stream outer loop + individual-axis row-block tile, host accumulators so the `(n_indiv, n_indiv)` output can be larger than GPU memory. `grm` runs two passes (chromosome-wide allele frequencies, then per-chunk standardized outer product).

**Library hardening**
- `HaplotypeMatrix.load_pop_file` writes `sample_sets` in block (ploidy-0 then ploidy-1) ordering rather than interleaved, fixing a latent pop-subset bug in eager `ibs` / `grm` where the diploid pairing in `_get_genotype_data` silently misaligned.
- `compute_counts_for_pairs` accepts a precomputed `has_missing` so the streaming LD path doesn't re-scan the stitched matrix per pair batch.
- `ZarrGenotypeSource._resolve_pop_file` accepts `sample_id` as a header alias alongside `sample`.

## Performance

Single-A100 chr15 scan (50 k diploids per population, 200 k haplotypes, 11.6 M variants):

| Stage | Wall |
|---|---|
| Per-window diversity + divergence at 10 kb / 100 kb / 1 Mb, marginal + joint SFS, per-pop Garud's H (one chromosome walk) | 905 s |
| LD decay (DD, Dz, π², σ_d²) at 9 bins to 200 kb across 16 probe regions (5 k-hap subsample/pop) | 52 s |
| Pairwise r² heatmap on a 1 Mb sub-region (5 k haps, 2 k SNPs after MAF filter) | 1 s |
| Total wall including figure rendering | ~16 min |

`slice_subsample_gpu` is the dominant win for sample-subset reads: 170 s → 1.8 s on a 5 Mb / 10 k-hap probe.

Validation: `ibs` and `grm` were verified against first-principles numpy on hand-built (4 indiv × 5 site) reference cases and against real Ag1000G 3R data (eager-vs-streaming match to machine precision: max abs diff 0.0 for IBS, 4e-13 for GRM). LD decay's unbiased DD estimator was verified against the formula on the same 2-variant hand-built case.

## Dependencies

`kvikio-cu12 >= 25.0` and `nvidia-nvcomp-cu12 >= 4.0` from `https://pypi.nvidia.com` are added to `pixi.toml`. The GPU-decode path is the default for bio2zarr-style stores; whole-sample-axis chunked stores fall back to the host fetcher with a one-time `BadlyChunkedWarning`.

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -17,6 +17,47 @@ GenotypeMatrix
    :undoc-members:
    :show-inheritance:
 
+Biobank-Scale Streaming
+-----------------------
+
+For VCZ stores that do not fit on the GPU eagerly,
+``HaplotypeMatrix.from_zarr`` / ``GenotypeMatrix.from_zarr`` can
+return a streaming view (``streaming='always'``, or
+``streaming='auto'`` with a too-large eager footprint). The streaming
+object iterates the chromosome chunk by chunk through every kernel
+that accepts an eager matrix; see :doc:`tutorials/biobank_streaming`
+for the end-to-end pattern.
+
+.. autoclass:: pg_gpu.streaming_matrix.StreamingHaplotypeMatrix
+   :members: iter_gpu_chunks, materialize, num_haplotypes, num_variants,
+             chrom, chrom_start, chrom_end, sample_sets, align_bp,
+             compute_ld_statistics_gpu_single_pop,
+             compute_ld_statistics_gpu_two_pops
+   :show-inheritance:
+
+.. autoclass:: pg_gpu.streaming_matrix.StreamingGenotypeMatrix
+   :members: iter_gpu_chunks, materialize, num_individuals, num_variants,
+             chrom, chrom_start, chrom_end, sample_sets, align_bp
+   :show-inheritance:
+
+.. autoclass:: pg_gpu.zarr_source.ZarrGenotypeSource
+   :members: slice_region, slice_region_gpu, slice_subsample,
+             slice_subsample_gpu, iter_chunks
+   :show-inheritance:
+
+.. autoclass:: pg_gpu.streaming_matrix.KvikioChunkFetcher
+   :members: iter_chunks, close
+   :show-inheritance:
+
+.. autoclass:: pg_gpu.streaming_matrix.HostChunkFetcher
+   :members: iter_chunks
+   :show-inheritance:
+
+.. autoclass:: pg_gpu.BiobankScaleWarning
+   :show-inheritance:
+
+.. autofunction:: pg_gpu.zarr_io.allel_zarr_to_vcz
+
 LD Statistics
 -------------
 
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -20,13 +20,16 @@ GenotypeMatrix
 Biobank-Scale Streaming
 -----------------------
 
-For VCZ stores that do not fit on the GPU eagerly,
-``HaplotypeMatrix.from_zarr`` / ``GenotypeMatrix.from_zarr`` can
-return a streaming view (``streaming='always'``, or
-``streaming='auto'`` with a too-large eager footprint). The streaming
-object iterates the chromosome chunk by chunk through every kernel
-that accepts an eager matrix; see :doc:`tutorials/biobank_streaming`
-for the end-to-end pattern.
+A *VCZ store* is bio2zarr's Zarr-on-disk encoding of a VCF: the
+genotype matrix is split into compressed chunks, each chunk a small
+array of samples by variants. For VCZ stores that do not fit
+entirely in GPU memory, ``HaplotypeMatrix.from_zarr`` /
+``GenotypeMatrix.from_zarr`` can return a streaming view
+(``streaming='always'``, or ``streaming='auto'`` with a too-large
+projected footprint). The streaming object iterates the chromosome
+chunk by chunk through every kernel that accepts a fully loaded
+matrix; see :doc:`tutorials/biobank_streaming` for the end-to-end
+pattern and the VCF-to-VCZ conversion step.
 
 .. autoclass:: pg_gpu.streaming_matrix.StreamingHaplotypeMatrix
    :members: iter_gpu_chunks, materialize, num_haplotypes, num_variants,
@@ -53,7 +56,7 @@ for the end-to-end pattern.
    :members: iter_chunks
    :show-inheritance:
 
-.. autoclass:: pg_gpu.BiobankScaleWarning
+.. autoclass:: pg_gpu.MemoryLimitedWarning
    :show-inheritance:
 
 .. autofunction:: pg_gpu.zarr_io.allel_zarr_to_vcz
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -206,6 +206,7 @@ Joint (Two-Population)
 .. autofunction:: pg_gpu.sfs.joint_sfs_folded
 .. autofunction:: pg_gpu.sfs.joint_sfs_scaled
 .. autofunction:: pg_gpu.sfs.joint_sfs_folded_scaled
+.. autofunction:: pg_gpu.sfs.project_joint_sfs
 
 Scaling and Folding Utilities
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -29,7 +29,8 @@ entirely in GPU memory, ``HaplotypeMatrix.from_zarr`` /
 projected footprint). The streaming object iterates the chromosome
 chunk by chunk through every kernel that accepts a fully loaded
 matrix; see :doc:`tutorials/biobank_streaming` for the end-to-end
-pattern and the VCF-to-VCZ conversion step.
+pattern and the VCF-to-VCZ conversion step. The reader requires
+diploid genotypes; haploid and polyploid stores are rejected.
 
 .. autoclass:: pg_gpu.streaming_matrix.StreamingHaplotypeMatrix
    :members: iter_gpu_chunks, materialize, num_haplotypes, num_variants,
--- a/docs/source/features.rst
+++ b/docs/source/features.rst
@@ -411,6 +411,35 @@ Distance Distribution Statistics
      - Variance, skewness, and kurtosis in one call
      - Schrider et al. (2018)
 
+Biobank-Scale Streaming
+-----------------------
+
+VCZ stores too large for the GPU eagerly (tens to hundreds of
+thousands of haplotypes) open via ``HaplotypeMatrix.from_zarr`` /
+``GenotypeMatrix.from_zarr`` as a streaming view that walks the
+chromosome chunk by chunk. Every per-chunk-reducible kernel below
+dispatches transparently on the streaming object -- the calling code
+is identical to the eager path. See :doc:`tutorials/biobank_streaming`.
+
+.. list-table::
+   :header-rows: 1
+   :widths: 35 65
+
+   * - Entry point
+     - Streaming behavior
+   * - ``windowed_analysis``
+     - Per-chunk dispatch + row-concat; window grids are aligned to the chunk grid.
+   * - ``sfs.sfs``, ``sfs.joint_sfs``, ``sfs.sfs_folded``, ``sfs.joint_sfs_folded``
+     - Per-chunk SFS bincount, summed across chunks.
+   * - ``HaplotypeMatrix.compute_ld_statistics_gpu_single_pop`` / ``_two_pops``
+     - Per-chunk pair-bin sums with a tail-buffer that captures pairs straddling chunk boundaries exactly once. Returns moments-LD DD, Dz, pi² (3 stats single-pop, 15 stats two-pop).
+   * - ``relatedness.ibs``, ``relatedness.grm``
+     - Variant-axis streamed; individual axis tiled into row blocks. ``(n_ind, n_ind)`` accumulators live on host so the output can exceed GPU memory. ``grm`` is two-pass (chromosome-wide allele frequencies, then per-chunk standardized outer product).
+   * - ``StreamingHaplotypeMatrix.materialize(region, sample_subset)``
+     - Sub-region eager build for kernels that need every variant simultaneously (``pairwise_r2``, Garud H, custom recipes). At biobank scale the sample-subset read decompresses directly to the GPU via kvikio + nvCOMP.
+   * - ``zarr_io.allel_zarr_to_vcz``
+     - Streaming converter from scikit-allel layout to VCZ for stores that pre-date bio2zarr.
+
 Fused Windowed Statistics
 -------------------------
 
--- a/docs/source/features.rst
+++ b/docs/source/features.rst
@@ -414,29 +414,44 @@ Distance Distribution Statistics
 Biobank-Scale Streaming
 -----------------------
 
-VCZ stores too large for the GPU eagerly (tens to hundreds of
-thousands of haplotypes) open via ``HaplotypeMatrix.from_zarr`` /
+A *VCZ store* is a Zarr-on-disk encoding of a VCF: the genotype
+matrix is split into compressed chunks, each chunk a small array of
+samples by variants. ``pg_gpu`` reads VCZ stores; if your data
+is in VCF you can convert it with the bio2zarr tools
+(``vcf2zarr explode`` then ``vcf2zarr encode``). The streaming
+codepath needs that VCZ layout because it relies on per-chunk
+decode -- a VCF row-text dump can't be sliced or decompressed
+chunk-wise the way zarr can. See
+:doc:`tutorials/biobank_streaming` for the VCF→VCZ conversion
+and a worked end-to-end example.
+
+A VCZ store too large for the GPU (tens to hundreds of thousands
+of haplotypes) opens via ``HaplotypeMatrix.from_zarr`` /
 ``GenotypeMatrix.from_zarr`` as a streaming view that walks the
-chromosome chunk by chunk. Every per-chunk-reducible kernel below
-dispatches transparently on the streaming object -- the calling code
-is identical to the eager path. See :doc:`tutorials/biobank_streaming`.
+chromosome chunk by chunk; every kernel listed below dispatches
+on the streaming object the same way it would on a fully loaded
+matrix -- the calling code is identical to the in-memory path.
+
+What runs on a streaming matrix:
 
 .. list-table::
    :header-rows: 1
    :widths: 35 65
 
    * - Entry point
-     - Streaming behavior
+     - How it streams
    * - ``windowed_analysis``
-     - Per-chunk dispatch + row-concat; window grids are aligned to the chunk grid.
+     - Each chunk's windows are computed on the GPU as the chunk arrives, then rows are concatenated; window grids are aligned to the chunk grid so no window straddles a chunk boundary.
    * - ``sfs.sfs``, ``sfs.joint_sfs``, ``sfs.sfs_folded``, ``sfs.joint_sfs_folded``
-     - Per-chunk SFS bincount, summed across chunks.
+     - Each chunk contributes its own per-site frequency-spectrum counts; the chromosome-wide answer is the sum across chunks.
+   * - ``sfs.project_joint_sfs``
+     - Same per-chunk accumulation, but the joint SFS is projected (via hypergeometric sampling) to a small target grid as it is built, so the full ``(n1+1, n2+1)`` histogram is never materialized.
    * - ``HaplotypeMatrix.compute_ld_statistics_gpu_single_pop`` / ``_two_pops``
-     - Per-chunk pair-bin sums with a tail-buffer that captures pairs straddling chunk boundaries exactly once. Returns moments-LD DD, Dz, pi² (3 stats single-pop, 15 stats two-pop).
+     - Variant-pair statistics within ``max_bp_dist`` are summed per chunk into the bp bins. Pairs that fall on opposite sides of a chunk boundary would be missed by naive per-chunk sums, so the last ``max_bp_dist`` of one chunk is carried forward and paired with the start of the next, counted exactly once. Returns moments-LD ``DD``, ``Dz``, ``pi²`` (3 stats single-pop, 15 stats two-pop).
    * - ``relatedness.ibs``, ``relatedness.grm``
-     - Variant-axis streamed; individual axis tiled into row blocks. ``(n_ind, n_ind)`` accumulators live on host so the output can exceed GPU memory. ``grm`` is two-pass (chromosome-wide allele frequencies, then per-chunk standardized outer product).
+     - Streamed along the variant axis; the individual axis is tiled into row blocks. ``(n_ind, n_ind)`` accumulators live on host so the output can exceed GPU memory. ``grm`` is a two-pass operation (first to calculate chromosome-wide allele frequencies, second to accumulate a per-chunk outer product).
    * - ``StreamingHaplotypeMatrix.materialize(region, sample_subset)``
-     - Sub-region eager build for kernels that need every variant simultaneously (``pairwise_r2``, Garud H, custom recipes). At biobank scale the sample-subset read decompresses directly to the GPU via kvikio + nvCOMP.
+     - Pulls one sub-region (and optionally a subset of haplotypes) of the chromosome into a fully-loaded matrix on the GPU for kernels that need every variant simultaneously -- ``pairwise_r2``, Garud's H, or any custom recipe. At hundreds of thousands of haplotypes the kvikio + nvCOMP path lets the subset read decompress directly into GPU memory rather than round-tripping the full sample axis through the host.
    * - ``zarr_io.allel_zarr_to_vcz``
      - Streaming converter from scikit-allel layout to VCZ for stores that pre-date bio2zarr.
 
--- a/docs/source/features.rst
+++ b/docs/source/features.rst
@@ -414,14 +414,13 @@ Distance Distribution Statistics
 Biobank-Scale Streaming
 -----------------------
 
-A *VCZ store* is a Zarr-on-disk encoding of a VCF: the genotype
+A *VCZ store* is a Zarr encoding of a VCF stored on disk: the genotype
 matrix is split into compressed chunks, each chunk a small array of
 samples by variants. ``pg_gpu`` reads VCZ stores; if your data
 is in VCF you can convert it with the bio2zarr tools
 (``vcf2zarr explode`` then ``vcf2zarr encode``). The streaming
-codepath needs that VCZ layout because it relies on per-chunk
-decode -- a VCF row-text dump can't be sliced or decompressed
-chunk-wise the way zarr can. See
+codepath needs that VCZ layout because it relies ona fast per-chunk
+decode, an opperation that can't be done on a VCF. See
 :doc:`tutorials/biobank_streaming` for the VCF→VCZ conversion
 and a worked end-to-end example.
 
@@ -451,7 +450,7 @@ What runs on a streaming matrix:
    * - ``relatedness.ibs``, ``relatedness.grm``
      - Streamed along the variant axis; the individual axis is tiled into row blocks. ``(n_ind, n_ind)`` accumulators live on host so the output can exceed GPU memory. ``grm`` is a two-pass operation (first to calculate chromosome-wide allele frequencies, second to accumulate a per-chunk outer product).
    * - ``StreamingHaplotypeMatrix.materialize(region, sample_subset)``
-     - Pulls one sub-region (and optionally a subset of haplotypes) of the chromosome into a fully-loaded matrix on the GPU for kernels that need every variant simultaneously -- ``pairwise_r2``, Garud's H, or any custom recipe. At hundreds of thousands of haplotypes the kvikio + nvCOMP path lets the subset read decompress directly into GPU memory rather than round-tripping the full sample axis through the host.
+     - Pulls one sub-region (and optionally a subset of haplotypes) of the chromosome into GPU memory for kernels that need every variant simultaneously -- ``pairwise_r2``, Garud's H, or any custom recipe. 
    * - ``zarr_io.allel_zarr_to_vcz``
      - Streaming converter from scikit-allel layout to VCZ for stores that pre-date bio2zarr.
 
--- a/docs/source/features.rst
+++ b/docs/source/features.rst
@@ -419,8 +419,8 @@ matrix is split into compressed chunks, each chunk a small array of
 samples by variants. ``pg_gpu`` reads VCZ stores; if your data
 is in VCF you can convert it with the bio2zarr tools
 (``vcf2zarr explode`` then ``vcf2zarr encode``). The streaming
-codepath needs that VCZ layout because it relies ona fast per-chunk
-decode, an opperation that can't be done on a VCF. See
+codepath needs that VCZ layout because it relies on a fast
+per-chunk decode, an operation that can't be done on a VCF. See
 :doc:`tutorials/biobank_streaming` for the VCF→VCZ conversion
 and a worked end-to-end example.
 
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -36,6 +36,7 @@ Key Features
 * **Multi-population analyses** with flexible population specification
 * **8 theta estimators and 4 neutrality tests** (pi, theta_w, theta_h, theta_l, eta1, eta1_star, minus_eta1, minus_eta1_star, Tajima's D, Fay-Wu's H, Zeng's E, DH)
 * **Validated against scikit-allel** -- 29 statistics verified at machine precision using real Ag1000G data
+* **Biobank-scale streaming** -- VCZ stores too large to fit on the GPU eagerly open as a streaming view that walks the chromosome chunk by chunk; every per-window / SFS / moments-LD / pairwise relatedness kernel dispatches transparently. See :doc:`tutorials/biobank_streaming`.
 
 Installation
 ------------
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -36,7 +36,7 @@ Key Features
 * **Multi-population analyses** with flexible population specification
 * **8 theta estimators and 4 neutrality tests** (pi, theta_w, theta_h, theta_l, eta1, eta1_star, minus_eta1, minus_eta1_star, Tajima's D, Fay-Wu's H, Zeng's E, DH)
 * **Validated against scikit-allel** -- 29 statistics verified at machine precision using real Ag1000G data
-* **Biobank-scale streaming** -- VCZ stores too large to fit on the GPU eagerly open as a streaming view that walks the chromosome chunk by chunk; every per-window / SFS / moments-LD / pairwise relatedness kernel dispatches transparently. See :doc:`tutorials/biobank_streaming`.
+* **Biobank-scale streaming** -- VCZ stores too large to fit on the GPU open as a streaming view that walks the chromosome chunk by chunk; every per-window / SFS / moments-LD / pairwise relatedness kernel dispatches transparently. See :doc:`tutorials/biobank_streaming`.
 
 Installation
 ------------
--- a/docs/source/tutorials/biobank_streaming.rst
+++ b/docs/source/tutorials/biobank_streaming.rst
@@ -93,11 +93,11 @@ functions:
    # Marginal SFS uses the full panel (1D output, fits anywhere).
    sfs_afr = sfs.sfs(stream, population="AFR")
 
-   # Joint SFS at biobank scale: the full (n1+1, n2+1) histogram is
-   # 80 GB at 100k haps per population, so use project_joint_sfs
-   # instead. It applies a hypergeometric projection per variant and
-   # accumulates straight into the small target grid -- every variant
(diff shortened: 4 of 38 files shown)


docs/source/tutorials/biobank_streaming.rst
```diff
@@ -0,0 +1,209 @@
...
+* the joint SFS,
+* Garud's H,
+* moments-LD pair bins (``DD``, ``Dz``, ``pi2``, derived
+  :math:`\sigma_D^2`),
+* identity by state (``ibs``) and the genetic relationship matrix
+  (``grm``)
+
+-- are handled automatically: you call the same function as on a
+small in-memory matrix and pg_gpu takes care of the chunk-by-chunk
+accumulation.
```
nspope: "takes care of the chunk by chunk accumulation" ==>
"takes care of the chunk by chunk accumulation, returning the aggregated (chromosome-wide) result. For example,
``` 
1-2 line code block 
```
would return `pi` over the entire chromosome, while being careful to only load as many variants into GPU memory as would fit."

docs/source/tutorials/biobank_streaming.rst
```diff
@@ -0,0 +1,209 @@
...
+
+For statistics that need every variant in scope at the same time
+(a pairwise-:math:`r^2` heatmap, a Garud's H hash table, a per-
+individual-pair distance matrix), call
+``streaming_matrix.materialize(region=..., sample_subset=...)`` to
+read a smaller piece -- a sub-region of the chromosome, restricted
+to a subset of haplotypes -- into a regular in-memory
+``HaplotypeMatrix`` and run the statistic on that.
```
nspope: Another concrete example (like in comment above) would help here

docs/source/tutorials/biobank_streaming.rst
```diff
@@ -0,0 +1,209 @@
...
+================================
+
+When a VCZ (bio2zarr) dataset has too many haplotypes to fit on the
+GPU all at once -- tens to hundreds of thousands --
+``HaplotypeMatrix.from_zarr`` (and ``GenotypeMatrix.from_zarr``)
+gives you a *streaming matrix* that reads the chromosome one chunk
+at a time. The streaming object works with every per-window, SFS,
+LD, and pairwise relatedness function in pg_gpu with no change to
+your code; GPU memory use scales with one chunk, not the whole
+chromosome.
```
nspope: Missing a closing TLDR on how this impacts the user, e.g.
"This only matters to the user when ..."

There's some missing context here, too --- what is a Zarr, what is a VCZ, why a VCF won't work. I think ideally we want the gist to be obvious to a naive reader from the first paragraph.

docs/source/tutorials/biobank_streaming.rst
```diff
@@ -0,0 +1,209 @@
...
+.. code-block:: python
+
+   from pg_gpu import HaplotypeMatrix, windowed_analysis, sfs
+   from pg_gpu.relatedness import ibs, grm
+
+   stream = HaplotypeMatrix.from_zarr(
+       "biobank/chr15.vcz",
+       streaming="always",
+       chunk_bp=500_000,         # genomic chunk size (default auto)
+       prefetch=1,               # read-ahead depth (0 disables)
```
nspope: 'prefetch' should either be explained or hidden here

docs/source/tutorials/biobank_streaming.rst
```diff
@@ -0,0 +1,209 @@
...
+   region = (50_000_000, 51_000_000)
+   subsample = list(stream.sample_sets["AFR"][:5_000])
+   region_hm = stream.materialize(region=region, sample_subset=subsample)
+
+   # Run any pairwise statistic on the regular in-memory matrix.
+   r2 = region_hm.pairwise_r2()
+
+``materialize(sample_subset=...)`` reads the (variants × subsample)
+genotype block directly onto the GPU when the store is set up for
+it (zstd / blosc / lz4 / deflate compressed bio2zarr chunks decoded
```
nspope: the parenthetical reads like gibberish here

docs/source/tutorials/biobank_streaming.rst
```diff
@@ -0,0 +1,209 @@
...
+through kvikio + nvCOMP). At biobank scale this is roughly 60×
+faster than the equivalent CPU-side path: a 5 Mb / 10,000-haplotype
+block drops from ~170 seconds to ~3 seconds.
+
+Converting a scikit-allel store
+-------------------------------
+
+Older biobank releases (e.g. the Ag1000G ``AgamP3.phased.zarr``)
```
nspope: Is "biobank releases" the right word here? Maybe "data releases"?

docs/source/tutorials/biobank_streaming.rst
```diff
@@ -0,0 +1,209 @@
...
+   allel_zarr_to_vcz(
+       "AgamP3.phased.zarr",                    # scikit-allel layout
+       "AgamP3.phased.3R.vcz",                  # output VCZ
+       contig="3R",                              # required for grouped allel stores
+       region="3R:1_000_000-2_000_000",          # optional sub-range
+       variant_chunk=10_000, sample_chunk=1_000,
+   )
+
+The resulting store is then opened the normal way through
+``HaplotypeMatrix.from_zarr``.
```
nspope, line 243: How does the "sample axis chunking" comment fit in here (if at all)? Is there an ready footgun by which someone could produce a not-sample-axis-chunked zarr? If not, I'd lean towards omitting that technical detail; and if so or consolidating it here (maybe rename the section to "Required Zarr format")

docs/source/tutorials/biobank_streaming.rst
```diff
@@ -0,0 +1,209 @@
...
+
+Worked example
+--------------
+
+A complete biobank-scale scan -- per-window diversity and
+divergence at three window sizes (10 kb, 100 kb, 1 Mb), marginal
+and joint SFS, Garud's H per population, moments-LD decay across
+probe regions, and a pairwise-:math:`r^2` heatmap of a 1 Mb
+sub-region -- ships as
+``06_simulated_genome_scan/scripts/genome_scan_ooa.py`` in the
```
nspope: I'm not sure you want to point-out to the paper repo like this (esp without a link). Seems better to include the example in pg_gpu repo directly.

docs/source/tutorials/biobank_streaming.rst
```diff
@@ -0,0 +1,209 @@
...
+
+* Your data already fits on the GPU in one piece. ``streaming='auto'``
+  will pick the in-memory path automatically; that path is faster
+  per call because there is no per-chunk overhead.
+* Your statistic needs every variant at once and the whole
+  chromosome is too big to load. Use ``materialize(region=...)`` to
+  read a smaller sub-region instead.
+* Your store is in scikit-allel layout rather than VCZ. Convert it
+  first with ``allel_zarr_to_vcz``; the streaming reader only
+  accepts VCZ.
```
nspope, line 291: What if your data is in a VCF? Would be good to show how to go from VCF to VCZ (and make it clear *why* we're using VCZ in the first place, VCF==bad at this scale). This seems like some missing context in this doc

docs/source/api.rst
```diff
@@ -17,6 +17,47 @@ GenotypeMatrix
    :undoc-members:
    :show-inheritance:
 
+Biobank-Scale Streaming
+-----------------------
+
+For VCZ stores that do not fit on the GPU eagerly,
```
nspope: same comment as elsewhere: "what is a VCZ?" Seems like some more context is needed (or could link out to a "why VCZ" subsection)

docs/source/features.rst
```diff
@@ -411,6 +411,35 @@ Distance Distribution Statistics
      - Variance, skewness, and kurtosis in one call
      - Schrider et al. (2018)
 
+Biobank-Scale Streaming
+-----------------------
+
+VCZ stores too large for the GPU eagerly (tens to hundreds of
```
nspope: Will a reader hitting this section ask "What is a VCZ?" --- maybe need some more context here (on Zarr vs VCF)

docs/source/features.rst
```diff
@@ -411,6 +411,35 @@ Distance Distribution Statistics
...
+   * - ``windowed_analysis``
+     - Per-chunk dispatch + row-concat; window grids are aligned to the chunk grid.
+   * - ``sfs.sfs``, ``sfs.joint_sfs``, ``sfs.sfs_folded``, ``sfs.joint_sfs_folded``
+     - Per-chunk SFS bincount, summed across chunks.
+   * - ``HaplotypeMatrix.compute_ld_statistics_gpu_single_pop`` / ``_two_pops``
+     - Per-chunk pair-bin sums with a tail-buffer that captures pairs straddling chunk boundaries exactly once. Returns moments-LD DD, Dz, pi² (3 stats single-pop, 15 stats two-pop).
+   * - ``relatedness.ibs``, ``relatedness.grm``
+     - Variant-axis streamed; individual axis tiled into row blocks. ``(n_ind, n_ind)`` accumulators live on host so the output can exceed GPU memory. ``grm`` is two-pass (chromosome-wide allele frequencies, then per-chunk standardized outer product).
+   * - ``StreamingHaplotypeMatrix.materialize(region, sample_subset)``
+     - Sub-region eager build for kernels that need every variant simultaneously (``pairwise_r2``, Garud H, custom recipes). At biobank scale the sample-subset read decompresses directly to the GPU via kvikio + nvCOMP.
```
nspope: jargon:
"sub-region eager build"
"sample-subset read"

docs/source/features.rst
```diff
@@ -411,6 +411,35 @@ Distance Distribution Statistics
...
+   * - Entry point
+     - Streaming behavior
+   * - ``windowed_analysis``
+     - Per-chunk dispatch + row-concat; window grids are aligned to the chunk grid.
+   * - ``sfs.sfs``, ``sfs.joint_sfs``, ``sfs.sfs_folded``, ``sfs.joint_sfs_folded``
+     - Per-chunk SFS bincount, summed across chunks.
+   * - ``HaplotypeMatrix.compute_ld_statistics_gpu_single_pop`` / ``_two_pops``
+     - Per-chunk pair-bin sums with a tail-buffer that captures pairs straddling chunk boundaries exactly once. Returns moments-LD DD, Dz, pi² (3 stats single-pop, 15 stats two-pop).
+   * - ``relatedness.ibs``, ``relatedness.grm``
+     - Variant-axis streamed; individual axis tiled into row blocks. ``(n_ind, n_ind)`` accumulators live on host so the output can exceed GPU memory. ``grm`` is two-pass (chromosome-wide allele frequencies, then per-chunk standardized outer product).
```
nspope:
```suggestion
     - Streamed along the variant axis; the individual axis is tiled into row blocks. ``(n_ind, n_ind)`` accumulators live on host so the output can exceed GPU memory. ``grm`` is a two-pass operation (first to calculate chromosome-wide allele frequencies, second to accumulate a per-chunk outer product).
```

docs/source/features.rst
```diff
@@ -411,6 +411,35 @@ Distance Distribution Statistics
...
+   :widths: 35 65
+
+   * - Entry point
+     - Streaming behavior
+   * - ``windowed_analysis``
+     - Per-chunk dispatch + row-concat; window grids are aligned to the chunk grid.
+   * - ``sfs.sfs``, ``sfs.joint_sfs``, ``sfs.sfs_folded``, ``sfs.joint_sfs_folded``
+     - Per-chunk SFS bincount, summed across chunks.
+   * - ``HaplotypeMatrix.compute_ld_statistics_gpu_single_pop`` / ``_two_pops``
+     - Per-chunk pair-bin sums with a tail-buffer that captures pairs straddling chunk boundaries exactly once. Returns moments-LD DD, Dz, pi² (3 stats single-pop, 15 stats two-pop).
```
nspope: "per-chunk pair-bin sums with a tail-buffer" == jargon

docs/source/features.rst
```diff
@@ -411,6 +411,35 @@ Distance Distribution Statistics
...
+``GenotypeMatrix.from_zarr`` as a streaming view that walks the
+chromosome chunk by chunk. Every per-chunk-reducible kernel below
+dispatches transparently on the streaming object -- the calling code
+is identical to the eager path. See :doc:`tutorials/biobank_streaming`.
+
+.. list-table::
+   :header-rows: 1
+   :widths: 35 65
+
+   * - Entry point
```
nspope, line 440: I think this list needs some humanizing

docs/source/index.rst
```diff
@@ -36,6 +36,7 @@ Key Features
 * **Multi-population analyses** with flexible population specification
 * **8 theta estimators and 4 neutrality tests** (pi, theta_w, theta_h, theta_l, eta1, eta1_star, minus_eta1, minus_eta1_star, Tajima's D, Fay-Wu's H, Zeng's E, DH)
 * **Validated against scikit-allel** -- 29 statistics verified at machine precision using real Ag1000G data
+* **Biobank-scale streaming** -- VCZ stores too large to fit on the GPU eagerly open as a streaming view that walks the chromosome chunk by chunk; every per-window / SFS / moments-LD / pairwise relatedness kernel dispatches transparently. See :doc:`tutorials/biobank_streaming`.
```
nspope:
```suggestion
* **Biobank-scale streaming** -- VCZ stores too large to fit on the GPU open as a streaming view that walks the chromosome chunk by chunk; every per-window / SFS / moments-LD / pairwise relatedness kernel dispatches transparently. See :doc:`tutorials/biobank_streaming`.
```

pg_gpu/__init__.py
```diff
@@ -63,7 +63,8 @@
     corners,
 )
 from .resampling import block_jackknife, block_bootstrap
+from ._biobank_warning import BiobankScaleWarning
```
nspope: I'm not sure I like naming this stuff "BiobankScaleWhatever" --- not because it's wrong, but because it's jargony and obscures the point. Wouldn't "TooMuchDataWarning" or "MemoryLimitedWarning" be better?

pg_gpu/_biobank_warning.py
```diff
@@ -0,0 +1,169 @@
+"""``BiobankScaleWarning`` and the size check ``from_vcf`` uses to fire it.
```
nspope: See comment above wrt naming. I think this is a jargony name that obscures the actual issue being flagged

nspope (commented):
Awesome --- here's comments on the docs additions. Will look over the implementation later

pg_gpu/_memutil.py
```diff
@@ -30,6 +30,27 @@ def estimate_variant_chunk_size(n_hap, bytes_per_element=4, n_intermediates=3,
...
+
+    Each row block holds ~``n_intermediates`` working arrays of shape
+    ``(block_size, n_ind)`` on the GPU (typically the row-block slice
+    of the indicator matmul plus the matmul output). Budgets the
+    requested fraction of free GPU memory for those.
+
+    Returns at least 1 and at most ``n_ind`` (a single block covers
+    every individual, equivalent to no tiling).
+    """
+    free = cp.cuda.Device().mem_info[0]
```
nspope, line 52: is this being computed on the fly? we have at most one thread doing the memory querying/bookkeeping, right? so there's no risk of a race condition with memory bookkeeping

pg_gpu/genotype_matrix.py
```diff
@@ -406,34 +413,116 @@ def from_zarr(cls, path, region=None, accessible_bed=None,
...
+            store and loads it if present. Pass ``False`` to disable
+            the auto-load.
+        streaming : {'auto', 'always', 'never'}, optional
+            Whether to return a ``StreamingGenotypeMatrix`` that
+            iterates the store chunk-by-chunk through the GPU
+            (suitable for biobank-scale stores that do not fit
+            eagerly on the device). ``'auto'`` (default) checks the
+            projected eager footprint against free GPU memory and
+            picks streaming when the eager matrix would consume more
+            than half the device. Scikit-allel layouts always route
```
nspope:
```suggestion
            than half the device. Scikit-allel-formatted zarr always route
```

pg_gpu/genotype_matrix.py
```diff
@@ -406,34 +413,116 @@ def from_zarr(cls, path, region=None, accessible_bed=None,
...
+            to eager because the streaming source is VCZ-only.
```
nspope:
```suggestion
            to eager because the only allowed format for 'streaming' is VCZ.
```

pg_gpu/genotype_matrix.py
```diff
@@ -406,34 +413,116 @@ def from_zarr(cls, path, region=None, accessible_bed=None,
...
+            format ``GenotypeMatrix.load_pop_file`` expects. Default
+            (``None``) looks for ``<path>.pops.tsv`` next to the
+            store and loads it if present. Pass ``False`` to disable
+            the auto-load.
+        streaming : {'auto', 'always', 'never'}, optional
+            Whether to return a ``StreamingGenotypeMatrix`` that
+            iterates the store chunk-by-chunk through the GPU
+            (suitable for biobank-scale stores that do not fit
+            eagerly on the device). ``'auto'`` (default) checks the
+            projected eager footprint against free GPU memory and
```
nspope: need to clarify what "eager" means here --- it's load everything into memory, right, no chunking? If so let's make that explicit. We can use eager in the internal comments/docstrings but let's make the user-facing stuff as clear as possible

pg_gpu/genotype_matrix.py
```diff
@@ -406,34 +413,116 @@ def from_zarr(cls, path, region=None, accessible_bed=None,
...
+            picks streaming when the eager matrix would consume more
+            than half the device. Scikit-allel layouts always route
+            to eager because the streaming source is VCZ-only.
+        chunk_bp, prefetch, backend
+            See ``HaplotypeMatrix.from_zarr``; ignored on the eager
```
nspope:
```suggestion
            See ``HaplotypeMatrix.from_zarr``; ignored when the data are loaded eagerly into memory
```

pg_gpu/genotype_matrix.py
```diff
@@ -406,34 +413,116 @@ def from_zarr(cls, path, region=None, accessible_bed=None,
...
+            (``None``) looks for ``<path>.pops.tsv`` next to the
+            store and loads it if present. Pass ``False`` to disable
+            the auto-load.
+        streaming : {'auto', 'always', 'never'}, optional
+            Whether to return a ``StreamingGenotypeMatrix`` that
+            iterates the store chunk-by-chunk through the GPU
+            (suitable for biobank-scale stores that do not fit
+            eagerly on the device). ``'auto'`` (default) checks the
+            projected eager footprint against free GPU memory and
+            picks streaming when the eager matrix would consume more
```
nspope:
```suggestion
            picks streaming when the full genotype matrix would consume more
```

pg_gpu/genotype_matrix.py
nspope:
```suggestion
            projected footprint of the genotype matrix against free GPU memory and
```

pg_gpu/genotype_matrix.py
```diff
@@ -406,34 +413,116 @@ def from_zarr(cls, path, region=None, accessible_bed=None,
...
+            Tab-delimited file mapping ``sample`` -> ``pop`` in the
+            format ``GenotypeMatrix.load_pop_file`` expects. Default
+            (``None``) looks for ``<path>.pops.tsv`` next to the
+            store and loads it if present. Pass ``False`` to disable
+            the auto-load.
+        streaming : {'auto', 'always', 'never'}, optional
+            Whether to return a ``StreamingGenotypeMatrix`` that
+            iterates the store chunk-by-chunk through the GPU
+            (suitable for biobank-scale stores that do not fit
+            eagerly on the device). ``'auto'`` (default) checks the
```
nspope:
```suggestion
            on the device). ``'auto'`` (default) checks the
```

pg_gpu/genotype_matrix.py
```diff
@@ -406,34 +413,116 @@ def from_zarr(cls, path, region=None, accessible_bed=None,
...
+        pop_file : str or False, optional
+            Tab-delimited file mapping ``sample`` -> ``pop`` in the
+            format ``GenotypeMatrix.load_pop_file`` expects. Default
+            (``None``) looks for ``<path>.pops.tsv`` next to the
+            store and loads it if present. Pass ``False`` to disable
+            the auto-load.
+        streaming : {'auto', 'always', 'never'}, optional
+            Whether to return a ``StreamingGenotypeMatrix`` that
+            iterates the store chunk-by-chunk through the GPU
+            (suitable for biobank-scale stores that do not fit
```
nspope:
```suggestion
            (suitable for large-scale stores that do not fit
```

pg_gpu/haplotype_matrix.py
```diff
@@ -7,6 +7,75 @@
...
+def _decide_streaming_mode(zarr_path, region, streaming, pop_file,
+                           free_gpu_bytes=None,
+                           fraction=STREAMING_AUTO_EAGER_FRACTION):
+    """Pick ``'eager'`` vs ``'streaming'`` based on the projected matrix size.
+
+    Returns ``(mode, source)`` where ``source`` is the
+    ``ZarrGenotypeSource`` opened for the size probe and reused
+    downstream (when ``mode == 'streaming'``), or ``None`` when no
+    source was opened (scikit-allel layout, missing store, or the
+    eager-fits branch where the caller's eager path opens its own
```
nspope: this sentence is a bit confusing even for an internal docstring--- what's the "eager-fits branch"?

pg_gpu/haplotype_matrix.py
```diff
@@ -7,6 +7,75 @@
...
+    store via ``read_genotypes``).
+
+    Scikit-allel layouts always return ``('eager', None)``: the
+    streaming source is VCZ-only, so the size check would refuse a
+    large allel store with nowhere to fall back to.
+
+    ``free_gpu_bytes`` is exposed for tests; production callers leave
```
nspope:
```suggestion
    ``free_gpu_bytes`` is exposed for tests; typical invocation will leave
```

pg_gpu/haplotype_matrix.py
```diff
@@ -7,6 +7,75 @@
...
+
+    Returns ``(mode, source)`` where ``source`` is the
+    ``ZarrGenotypeSource`` opened for the size probe and reused
+    downstream (when ``mode == 'streaming'``), or ``None`` when no
+    source was opened (scikit-allel layout, missing store, or the
+    eager-fits branch where the caller's eager path opens its own
+    store via ``read_genotypes``).
+
+    Scikit-allel layouts always return ``('eager', None)``: the
+    streaming source is VCZ-only, so the size check would refuse a
```
nspope:
```suggestion
    streaming format is VCZ-only, so the size check would refuse a
```

pg_gpu/haplotype_matrix.py
```diff
@@ -7,6 +7,75 @@
...
+
+    ``free_gpu_bytes`` is exposed for tests; production callers leave
+    it None to let ``cp.cuda.Device().mem_info`` provide it.
+    """
+    import zarr
+    from .zarr_io import detect_zarr_layout
+    try:
+        store = zarr.open_group(zarr_path, mode="r")
+        layout = detect_zarr_layout(store)
+    except (FileNotFoundError, KeyError, ValueError):
```
nspope, line 43: the idea here is that we'll error out, but are deferring erroring out until a helper function encounters the malformed path?

pg_gpu/haplotype_matrix.py
```diff
@@ -7,6 +7,75 @@
...
+
+    if free_gpu_bytes is None:
+        import cupy as _cp
+        free_gpu_bytes = int(_cp.cuda.Device().mem_info[0])
+
+    if eager_bytes <= fraction * free_gpu_bytes:
+        return "eager", None
+    if streaming == "never":
+        raise MemoryError(
+            f"streaming='never' but the eager matrix would be "
```
nspope:
```suggestion
            f"streaming='never' but the full matrix would be "
```

pg_gpu/haplotype_matrix.py
```diff
@@ -375,33 +451,135 @@ def from_zarr(cls, path: str, region: str = None,
             Genomic region 'chrom:start-end' to load a subset.
         accessible_bed : str, optional
             Path to a BED file defining accessible/callable regions.
+        pop_file : str or False, optional
```
nspope: it would be nice to allow a zarr key to be passed here, if the population info is already in the zarr (i.e. a non-required field in the VCZ). Or, a numpy array. It's pretty cumbersome to have to write out metadata in a particular format.

pg_gpu/haplotype_matrix.py
```diff
@@ -375,33 +451,135 @@ def from_zarr(cls, path: str, region: str = None,
...
+            Tab-delimited file mapping ``sample`` -> ``pop`` in the
+            format ``HaplotypeMatrix.load_pop_file`` expects. Default
+            (``None``) looks for ``<path>.pops.tsv`` next to the
+            store and loads it if present, announcing the auto-load
+            to stderr. Pass ``False`` to disable the auto-load.
+        streaming : {'auto', 'always', 'never'}, optional
+            Whether to return a ``StreamingHaplotypeMatrix`` that
+            iterates the store chunk-by-chunk through the GPU
+            (suitable for biobank-scale stores that do not fit
+            eagerly on the device). ``'always'`` forces streaming;
```
nspope: Edit this for parity with the equivalent docstring for GenotypeMatrix (see comments above on "eager" in user-facing docstrings)

nspope: Maybe should just point GenotypeMatrix docstring to HaplotypeMatrix docstring, to avoid having to write the same docs in two places?

pg_gpu/haplotype_matrix.py
```diff
@@ -375,33 +451,135 @@ def from_zarr(cls, path: str, region: str = None,
...
 
         data = read_genotypes(path, region)
         gt = data['gt']
         positions = data['positions']
         sample_names = data['samples']
 
-        num_variants, num_samples, ploidy = gt.shape
-        assert ploidy == 2
+        if gt.shape[2] != 2:
+            raise ValueError(f"expected ploidy 2; got {gt.shape[2]}")
```
nspope: what happens if the data are haploid?

pg_gpu/streaming_matrix.py
```diff
@@ -0,0 +1,679 @@
...
+
+class BadlyChunkedWarning(UserWarning):
+    """Emitted when ``backend='auto'`` picks ``host`` on a store whose
+    call_genotype chunking would have defeated the kvikio fetcher's
+    win. The store is functional but a bio2zarr-style re-encode would
+    unlock the GPU-decode fast path."""
+
+
+def _pick_chunk_fetcher(source, *, backend):
+    """Pick a fetcher for ``source``. Errors and warnings reflect what the
```
nspope, line 108: What are the hardware requirements for kvikio, if any? We should document these up front

pg_gpu/streaming_matrix.py
```diff
@@ -0,0 +1,679 @@
...
+class KvikioChunkFetcher(ChunkFetcher):
+    """Read chunks via ``kvikio.zarr.GDSStore`` with on-GPU codec decode.
+
+    Reads the store through ``kvikio.zarr.GDSStore`` and enables
+    zarr 3's GPU buffer prototype so nvCOMP decodes the codec on the
+    GPU. The chunk array therefore lands directly on the device --
+    ``build_haplotype_matrix`` skips the host-to-device copy in
+    ``cp.asarray(gt)`` because ``gt`` is already a cupy array.
+
+    The win is in the GPU codec decode, not in GPU Direct Storage.
```
nspope:
```suggestion
    The crucial performance gain is in the GPU-accelerated codec decode, not in GPU Direct Storage.
```

pg_gpu/streaming_matrix.py
```diff
@@ -0,0 +1,679 @@
...
+    By default this fetcher sets ``kvikio.defaults.compat_mode=ON``
+    so kvikio reads bytes through posix into bounce buffers without
+    trying to negotiate a real GDS handshake (which is slower on
```
nspope: what is a "real GDS handshake" --- can we rewrite this to be more readable for future-us

pg_gpu/windowed_analysis.py
```diff
@@ -1061,6 +1061,101 @@ def scatter_sum(values):
...
+        raise NotImplementedError(
+            "local_pca requires a chromosome-wide reference and is not "
+            "available on the StreamingHaplotypeMatrix path; materialize "
+            "the region eagerly to run it."
+        )
+    _garud_stats = {"garud_h1", "garud_h12", "garud_h123", "garud_h2h1",
+                    "garud_n_distinct", "haplotype_count"}
+    if any(s in _garud_stats for s in statistics):
+        raise NotImplementedError(
+            "Garud H per-chunk dispatch is correctly basis-stable now "
```
nspope: this error message is overkill; let's say something like "this is not implemented" in the error message, and leave the technical detail to a comment

pg_gpu/windowed_analysis.py
```diff
@@ -2390,6 +2496,42 @@ def windowed_statistics_fused_chunked(haplotype_matrix: HaplotypeMatrix,
     return results
 
 
+def _position_weights(positions, salt):
+    """Position-deterministic float64 weights for the Garud H hash basis.
+
+    Replaces the seed=42 RNG draw of length n_var. That draw produced a
```
nspope: do we need to document what this replaced?

pg_gpu/zarr_io.py
```diff
@@ -13,6 +14,27 @@ def _parse_region(region):
     return chrom, start, end
 
 
+def resolve_pop_file_path(zarr_path, pop_file, *, announce_prefix):
```
nspope, line 17: see comment above about allowing a Zarr key here (or numpy array/list) instead

pg_gpu/zarr_io.py
```diff
@@ -190,6 +232,143 @@ def write_vcz(zarr_path, gt, positions, samples=None, contig_name=None):
                            data=np.zeros(len(positions), dtype=np.int8))
 
 
+def allel_zarr_to_vcz(allel_path, vcz_path, *, contig=None, region=None,
```
nspope, line 332: hmmm ... feels like there should be a schema set (or check) here: for example, what if vcz schema changes in bio2zarr in the future; can we automatically detect that we're writing invalid output? maybe this check (if possible) would belong in the test suite?

nspope (commented):
Comments on implementation (mostly docstrings), still need to glance over unit tests. Only major "gap" is that it'd be nice to pass in a numpy array or a zarr key name to fetch the population information, rather than relying on writing an external metadata file (which seems unnecessarily clunky). We can do this in a followup PR if you'd like (open a new issue if you agree on this feature)

tests/test_streaming.py
```diff
@@ -0,0 +1,356 @@
...
+class TestStreamingEquivalence:
+
+    def test_stream_concat_matches_eager(self, vcz_store):
+        # walking every chunk in streaming mode and concatenating the
+        # per-chunk haplotype matrices reproduces the eager matrix
+        # bit-for-bit -- the invariant streaming-aware kernels rely on.
+        path, _ = vcz_store
+        eager = HaplotypeMatrix.from_zarr(path, streaming="never")
+        smatrix = HaplotypeMatrix.from_zarr(path, streaming="always",
+                                            chunk_bp=5_000, prefetch=1)
```
nspope: maybe pytest mark parametrize "chunk_bp" and "prefetch" here so we get some range of behaviours for streaming

tests/test_streaming_ld.py
```diff
@@ -0,0 +1,204 @@
...
+
+class TestSinglePopParity:
+
+    @pytest.mark.parametrize("chunk_bp", [50_000, 25_000])
+    def test_parity_single_chunk(self, vcz_store, chunk_bp):
+        # Bin upper edge is short relative to chunk width, so no pair
+        # crosses a chunk boundary -- the tail buffer should be empty
+        # at every chunk and the result must match eager exactly.
+        eager, stream = _aligned_single_pop_pair(vcz_store, chunk_bp=chunk_bp)
+        bp_bins = [0, 1_000, 5_000]  # max_dist 5kb << 25kb chunk
```
nspope, line 116: do we expect errors or poor behaviour if the streaming chunk size is less than window size in bp bins?

tests/test_streaming_kernels.py
```diff
@@ -0,0 +1,274 @@
...
+    eager = HaplotypeMatrix.from_zarr(vcz_store, streaming="never",
+                                      **stream_kwargs.get("from_zarr_kwargs", {}))
+    stream = HaplotypeMatrix.from_zarr(vcz_store, streaming="always",
+                                       **stream_kwargs)
+    eager.chrom_start = stream.chrom_start
+    eager.chrom_end = stream.chrom_end
+    return eager, stream
+
+
+class TestWindowedAnalysisDispatch:
```
nspope, line 81: do we expect any interaction between chunk bp size and window size? If so (or even if not) it might be good to pytest mark parametrize so we hit cases where chunk << window

nspope (commented):
Tests generally look good to me (I'm not following all the details, but the scope seems fine). Only comment is, it might be good to pytest mark parameterize things like chunk size and window size, to get coverage of any cases where weird interactions might happen.

andrewkern:
okay @nspope -- I've tried to go through and hit all your review points. take another look

pg_gpu/haplotype_matrix.py
```diff
@@ -375,33 +446,166 @@ def from_zarr(cls, path: str, region: str = None,
             Genomic region 'chrom:start-end' to load a subset.
         accessible_bed : str, optional
             Path to a BED file defining accessible/callable regions.
+        pop_file : str, numpy.ndarray, list, dict, or False, optional
```
nspope, line 449: minor API comment (feel free to disregard, but now is the time to change it): this might be better named `pop_assignment` now that it is multi-type.

pg_gpu/haplotype_matrix.py
```diff
@@ -375,33 +446,166 @@ def from_zarr(cls, path: str, region: str = None,
...
-        haplotypes[:, :num_samples] = gt[:, :, 0]
-        haplotypes[:, num_samples:] = gt[:, :, 1]
-        haplotypes = haplotypes.T
+        if gt.shape[2] != 2:
+            # The eager and streaming readers, the kvikio sample-subset
+            # gather, and the (n_dip, 2) -> 2 * n_dip haplotype layout all
+            # assume diploid. Haploid / polyploid stores need a different
+            # gather path and are not currently supported.
+            raise ValueError(
+                f"HaplotypeMatrix.from_zarr requires diploid (ploidy 2) "
```
nspope, line 556: Just a reminder to note this somewhere in the docs as well

nspope (approved):
Looks good to me --- one optional API nitpick

PR #109, opened 2026-05-19
feat: VCF/VCZ quality fields, filtering, and clean-VCZ output

### LINKED ISSUE #108: Expose VCF/VCZ quality metrics in the reader + filter-and-encode in vcf_to_zarr
## Motivation

A common per-genotype filter (e.g. `FMT/GQ >= 20 && FMT/DP >= 10`) on a VCF currently means leaving pg_gpu, running bcftools (or scikit-allel), and either feeding the filtered VCF back through `vcf_to_zarr` or coming in via a hand-curated VCZ. Once a VCZ has been built, the FORMAT arrays are sitting right there on disk (bio2zarr keeps them by default) but pg_gpu's reader ignores them, so the only way to act on them is round-trip through the source VCF again.

For many workflows on RAD-seq / low-coverage WGS this is a bottleneck: every threshold sweep means another bcftools pass over a multi-GB VCF.

## Current state

- `pg_gpu.zarr_io.vcf_to_zarr` is a thin wrapper over `bio2zarr.vcf.explode/encode`, which already preserves every FORMAT and INFO field present in the source VCF (e.g. `call_GQ`, `call_DP`, `call_AD`, `variant_AC`, `variant_AF`, ...). The VCZs we ship via this path *do* contain those arrays.
- `pg_gpu.zarr_io.read_genotypes_vcz` only pulls `call_genotype`, `variant_position`, `sample_id`. The FORMAT arrays are inaccessible through the standard read path.
- `HaplotypeMatrix.from_vcf` / `GenotypeMatrix.from_vcf` likewise only read `calldata/GT` + positions + samples; the rest of the `allel.read_vcf` payload is discarded.

So the data is preserved end-to-end at the VCZ level, but no filter primitive exists at any layer above raw zarr.

## Proposed surface

Four operations, ideally all in one PR or coordinated set:

1. **Read quality metrics from VCF.** `HaplotypeMatrix.from_vcf` / `GenotypeMatrix.from_vcf` accept a list of FORMAT/INFO fields to surface alongside `calldata/GT` (e.g. `format_fields=['GQ', 'DP']`). The fields land as numpy arrays accessible on the returned matrix.

2. **Filter the VCF itself.** A `vcf_filter(...)` (name TBD) entry point that takes a source VCF, a per-genotype predicate (e.g. `min_gq`, `min_dp`, or a callable), and either:
     - writes a filtered VCF, or
     - writes a filtered VCZ directly (skipping the intermediate VCF text), or
     - both, controlled by a kwarg.

3. **Write filtered sites to VCZ.** Folded into (2) — the common path is "filter the VCF, encode the result to VCZ". Should be a single call rather than two-step.

4. **Read quality metrics from VCZ.** `HaplotypeMatrix.from_zarr` / `GenotypeMatrix.from_zarr` accept `min_gq=`, `min_dp=` (and/or a general predicate over FORMAT arrays) and apply the mask at read time — set `call_genotype = -1` where the predicate fails. Also expose the underlying `call_GQ`/`call_DP` arrays for callers that want to do their own filtering. Threshold lives in code, not in the encoded data, so re-running with different cutoffs doesn't re-encode the VCF.

(4) is the cheap, high-value one — encoding is the expensive step, so making the cutoffs read-time means iterating on filters is essentially free.

## Notes on existing plumbing

- The mask machinery for `accessible_bed=` (`HaplotypeMatrix.set_accessible_mask` etc.) is the obvious template — same kwarg surface, same lazy-apply behavior.
- bio2zarr's `encode` step is field-aware; pruning or selecting which FORMAT fields to land in the VCZ is already a supported option there.
- Per-genotype masking should probably write to `call_genotype_mask` (already present in the VCZ schema) as well as setting `call_genotype = -1`, so a downstream reader using either convention sees the same result.

andrewkern:
Closes #108.

Adds a minimal four-surface addition for VCF FORMAT/INFO quality metrics. 
## API

```python

# 1. Load with quality fields (auto-detected as INFO or FORMAT by shape)
hm = HaplotypeMatrix.from_vcf("foo.vcf.gz", fields=["MQ", "GQ", "DP"])
hm = HaplotypeMatrix.from_zarr("foo.vcz", fields=["MQ", "GQ", "DP"])

# 2. Inspect: one dict, shape tells you per-variant vs per-genotype
hm.fields["MQ"]   # (n_var,)             -- INFO
hm.fields["GQ"]   # (n_var, n_samples)   -- FORMAT
# plt.hist(hm.fields["GQ"].ravel(), bins=50)  # raw arrays, user plots

# 3. Filter (returns a new matrix)
hm_clean = hm.filter(
    variants=hm.fields["MQ"] >= 40.0,
    genotypes=(hm.fields["GQ"] >= 20) & (hm.fields["DP"] >= 10),
    drop_all_missing=True,
)

# 4. Write clean VCZ -- round-trips surviving fields
hm_clean.to_zarr("clean.vcz", format="vcz")
```

Same surface on `GenotypeMatrix`. One new kwarg (`fields=`), one new attribute (`hm.fields`), one new method (`hm.filter`), one extended writer kwarg. Bare VCF tags throughout (`'MQ'`, `'GQ'`) -- internal layout names (`variants/MQ`, `calldata/GQ`, `variant_MQ`, `call_MQ`) stay in the I/O layer.

## Commits

1. **`4c725a2` feat: read VCF FORMAT/INFO quality fields via `fields=` kwarg.** Read path for all four layouts: VCF (via `allel.read_vcf` with header probing through `allel.read_vcf_headers`), VCZ (`variant_<tag>` / `call_<tag>`), scikit-allel flat (`variants/<tag>` / `calldata/<tag>`), scikit-allel grouped (same under the chrom subgroup). Region-restricted reads slice the QC arrays consistently with the genotype matrix via the `variant_indices` each genotype reader now returns. Missing tags warn and drop.
2. **`25ed8b6` feat: `filter()` method for variant + genotype QC masks.** Eager, copy semantics. `variants=` drops rows, `genotypes=` sets cells to `-1` (both haplotype rows on the haplotype-matrix side). `drop_all_missing=True` by default. Empty results route through the same `object.__new__` workaround `get_subset` uses so a too-aggressive filter returns a structured empty matrix instead of raising from the constructor.
3. **`2ffb68b` feat: round-trip fields through `write_vcz` / `to_zarr`.** `write_vcz(..., fields=...)` creates `variant_<tag>` (ndim=1) or `call_<tag>` (ndim=2) arrays in the VCZ output using the input dtype, so `load -> filter -> save -> reload` is byte-exact on every surviving QC array. Mismatched shapes raise before write so the output never disagrees with `call_genotype`. `format='scikit-allel'` with non-empty `fields` raises `NotImplementedError` to avoid silent drops.

## Tests

`tests/test_qc_fields.py` covers 26 cases: read paths (all four layouts), region alignment, missing-field warnings, the streaming guard, all combinations of `filter()` masks plus `drop_all_missing`, no-args copy, empty result, shape-mismatch guards, and three round-trip cases (filter-then-save preserves values, dtype is preserved, write_vcz rejects mismatched shapes).

Full repo `pytest tests/` (excluding the docs-example runners): **946 passed, 55 skipped, 0 failures**.

## Out of scope (deferred follow-ups)

* **Streaming filter on `StreamingHaplotypeMatrix` / `StreamingGenotypeMatrix`.** Combining `streaming=` with `fields=` raises `NotImplementedError`. Streaming filter is a separate piece of work that needs `ZarrGenotypeSource.slice_region` to pull extra arrays per chunk.
* **Filtering at VCZ-build time** inside `vcf_to_zarr` (bio2zarr-side per-genotype masking / FORMAT pruning during encode). Read-time path covers the common case -- encode once, iterate on thresholds.
* **Predicate sugar** (`hm.filter(min_gq=20, min_dp=10)`). Convenience kwargs are mechanical to add on top of the mask-based path.
* **`pg_gpu.moments_ld.compute_ld_statistics`** keeps reading VCFs through its own parser; no change in this branch.

### DIFF
--- a/docs/source/examples.rst
+++ b/docs/source/examples.rst
@@ -239,6 +239,42 @@ Compute multiple statistics across thousands of windows without Python loops:
 For end-to-end command-line workflows (``vcf_to_zarr.py``,
 ``genome_scan.py``), see :doc:`workflows`.
 
+Quality-Aware Filtering
+-----------------------
+
+Load VCF / VCZ FORMAT and INFO arrays alongside the genotype matrix,
+build per-variant and per-genotype masks from them, and write the
+filtered survivors back as a clean VCZ that round-trips the same
+arrays. The full walk-through is in
+:doc:`tutorials/qc_fields`; the recipe in one block:
+
+.. code-block:: python
+
+   from pg_gpu import HaplotypeMatrix
+
+   # Bare VCF tags; pg_gpu auto-resolves INFO vs FORMAT from the source.
+   hm = HaplotypeMatrix.from_vcf("cohort.vcf.gz",
+                                  fields=["MQ", "QD", "GQ", "DP"])
+   # or, equivalently, from a bio2zarr-encoded VCZ:
+   # hm = HaplotypeMatrix.from_zarr("cohort.vcz",
+   #                                 fields=["MQ", "QD", "GQ", "DP"])
+
+   # Shape disambiguates per-variant from per-genotype.
+   hm.fields["MQ"]   # (n_var,)             — INFO
+   hm.fields["GQ"]   # (n_var, n_samples)   — FORMAT
+   # Plot directly: plt.hist(hm.fields["GQ"].ravel(), bins=50)
+
+   # Mask-based filter. ``variants`` drops rows; ``genotypes`` sets
+   # per-cell calls to -1 (both allele rows on the haplotype side).
+   hm_clean = hm.filter(
+       variants=(hm.fields["MQ"] >= 40.0) & (hm.fields["QD"] >= 2.0),
+       genotypes=(hm.fields["GQ"] >= 20) & (hm.fields["DP"] >= 10),
+       drop_all_missing=True,
+   )
+
+   # Persist. The surviving QC arrays round-trip into the new store.
+   hm_clean.to_zarr("cohort.clean.vcz", format="vcz", contig_name="chr1")
+
 Missing Data
 ------------
 
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -33,6 +33,7 @@ Key Features
 * **Comprehensive statistics**: LD (D, D-squared, Dz, pi2, r/r-squared), diversity (pi, theta, Tajima's D, heterozygosity, Fay & Wu's H), divergence (FST Hudson/Weir-Cockerham/Nei, Dxy, Da, Snn, Gmin, dd, dd_rank, Zx), selection scans (iHS, XP-EHH, nSL, XP-nSL, Garud's H, EHH decay), SFS (unfolded, folded, joint, scaled), admixture (Patterson's F2, F3, D)
 * **Fused windowed analysis**: compute all statistics across all genomic windows in a single GPU pass -- up to 60x faster than scikit-allel
 * **Automatic missing data handling** across all modules
+* **Quality-aware filtering** -- load VCF FORMAT / INFO arrays (``GQ``, ``DP``, ``MQ``, ...) with ``fields=``, mask variants and genotypes from them, and round-trip the survivors into a clean VCZ. See :doc:`tutorials/qc_fields`.
 * **Multi-population analyses** with flexible population specification
 * **8 theta estimators and 4 neutrality tests** (pi, theta_w, theta_h, theta_l, eta1, eta1_star, minus_eta1, minus_eta1_star, Tajima's D, Fay-Wu's H, Zeng's E, DH)
 * **Validated against scikit-allel** -- 29 statistics verified at machine precision using real Ag1000G data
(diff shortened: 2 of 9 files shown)


andrewkern:
@stsmall take a look

stsmall:
@andrewkern this looks good. the below code gets everything without error.
```
REQUESTED = ['DP', 'GQ', 'MQ', 'AC', 'AN', 'ExcessHet', 'InbreedingCoeff', 'MLEAF', 'quality']
with warnings.catch_warnings(record=True) as wlist:
    warnings.simplefilter('always')
    hm = HaplotypeMatrix.from_zarr(
        ZARR, region='3L:1-2000000', fields=REQUESTED, streaming='never',
    )
```

PR #111, opened 2026-05-25
Add .claude/SKILL.md — Claude Code skill for pg_gpu

stsmall:
## Summary

Adds a single-file Claude Code skill at `.claude/SKILL.md` covering the full pg_gpu workflow: data loading (eager + streaming + `fields=`), accessibility masking, missing-data modes, per-population diversity / SFS / divergence / selection / 2-locus LD stats, windowed analyses, PCA, and block-jackknife.

Drop it into `~/.claude/skills/pg_gpu/` (or any path Claude Code is configured to scan) and the agent will follow its recipes when users ask about pg_gpu.

Verified end-to-end against the Ag1000G Ag3.0 X chromosome (~10 Mb, 1,470 diploids) — every documented stat family runs without modification.

## Test plan
- [ ] Confirm the file lands at `.claude/SKILL.md` cleanly
- [ ] Copy to `~/.claude/skills/pg_gpu/SKILL.md` and verify Claude Code auto-discovers the skill
- [ ] Optional: run the companion demo notebook (forthcoming separate contribution) against a current pg_gpu install

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #112, opened 2026-05-25
docs: add end-to-end skill walkthrough tutorial

stsmall:
## Summary

Adds a single RST tutorial — `docs/source/tutorials/pg_gpu_skill_walkthrough.rst` — that walks through the full pg_gpu pipeline end to end, mirroring the natural-language Claude Code skill (`.claude/SKILL.md`, added in #111).

Sections:

1. Load (`vcf_to_zarr` + `from_zarr(streaming='auto')`)
2. QC field inspection via direct zarr (CPU, no GPU memory cap)
3. Windowed π in three missingness modes (`include`, `exclude`, `pairwise`)
4. Accessibility-mask normalization (per-callable-base π)
5. PCA (biallelic + MAF > 0.05 + LD-prune + `randomized_pca` + metadata coloring)
6. Per-population diversity, SFS, pairwise FST/dxy, windowed FST/dxy
7. Selection scans (iHS, Garud's H)
8. 2-locus LD statistics via `moments_ld.compute_ld_statistics`
9. Block-jackknife confidence interval on π

Indexed under "Available tutorials" in `docs/source/tutorials.rst`.

Verified end-to-end on the Ag1000G Ag3.0 X chromosome (1,470 *A. gambiae* diploids, ~7.6 M variants on the first 10 Mb). Total runtime ~30 minutes on a single A100.

## Test plan
- [ ] `pixi run sphinx-build docs/source docs/build` — confirm the new RST renders without warnings
- [ ] Spot-check inline code blocks against the current pg_gpu API
- [ ] Optional: copy the snippets into a fresh Python session and rerun against your own VCZ

🤖 Generated with [Claude Code](https://claude.com/claude-code)


nspope:
Rad--- just to clarify, is this meant for a human to read or for Claude? If the latter, should it just live in the skill?

The intro paragraph says 

> Each section starts with a natural-language request a user might make of the Claude Code skill (``.claude/SKILL.md``); the code beneath is what the skill runs against pg_gpu's API.

and makes me think it's written for a human user. But it doesn't really seem written this way? That is, I'm not seeing the natural language prompts?

stsmall:
this is for a human to read. PR111 is the claude skill. Really very simple ... type "load my vcf" then an example of the code it assoc w/ that command and a plot. More useful for anyone not used to Claude.

nspope:
Yup, I understand... but where's the example natural language prompt per code block, that is advertised in the first paragraph? Is it just the section header? So for example is `QC field inspection — direct zarr (CPU-only, no memory cap)` intended to be the natural language prompt?

Same goes for "Windowed pi in 10kb windows, three missingness modes" --- this seems like a description of content, not a prompt.

Sorry if I'm being dense here!

andrewkern:
I agree with @nspope  here-- the framing is current off. it says:

>   Each section starts with a natural-language request a user might make of the Claude Code skill; the code beneath is what the skill runs against pg_gpu's API.

  but the section headers are just technical labels ("Load (VCZ if available, otherwise convert)", "Windowed π in 10 kb windows, two missingness modes"). there's no user voice anywhere. it reads as a normal API tutorial, not a skill walkthrough.
if we want the framing to land, each section needs a quoted prompt at the top — the thing a user would actually type at the skill — and then the code shown as what the skill turned that into. something like:

  3. "Run pi across 10 kb windows and compare missing-data modes"
  ---------------------------------------------------------------

  The skill picks ``windowed_analysis`` and loops over the two
  implemented modes: ...

I pushed an edit pass for correctness as some of the stuff was just factually not matched to the API

stsmall:
yeah that is how it was in my ipynb that I worked out, but somehow when moving to the rst that got lost. I will restore those and fix

stsmall:
I changed it, now there are user prompts, the corresponding skill interpretation, and then the code block.
Screenshot 2026-05-25 at 19 36 03
Screenshot 2026-05-25 at 19 37 58

andrewkern:
looks good to me.
 @nspope what do you think?

nspope:
Looks fine by me

andrewkern:
okay i will merge

PR #114, opened 2026-05-28
ts examples

attrna:
I added genome_scan_ts.py for tree sequences and two ts simulation scripts.

### One population with selection 
```
pixi add tszip
pixi run python examples/simulate_ts_selection_1pop.py --popsize 10000 --s 0.1 --samples 100 --output examples/ts_sel_1pop.tsz
pixi run python utils/genome_scan_ts.py examples/ts_sel_1pop.tsz -o examples/ts_sel_1pop_plot.pdf
```

### Two populations with split and optional migration
```
pixi run python examples/simulate_ts_2pop.py --output examples/ts_2pop.tsz
pixi run python utils/genome_scan_ts.py examples/ts_2pop.tsz -o examples/ts_2pop_plot.pdf
```


andrewkern:
Thanks @attrna !

PR #116, opened 2026-06-05
Make pg_gpu pip-installable from PyPI

### LINKED ISSUE #115: Integrating pg-gpu into existing workflows
I want to be able to use pg_gpu with my existing workflows, which include e.g. dependency management with conda.

How to use pg_gpu in a Snakemake rule or a Jupyter Notebook that imports it. Pixi is project specific and wouldn't mesh well with current set ups...

andrewkern:
Closes #115.

The easiest fix here is to copy our pixi env info (`pixi.toml`) to `pyproject.toml`. Currently `pyproject.toml` declared no dependencies at all -- every runtime requirement lived only in `pixi.toml` (by design), so a pip install produced a package that couldn't import.

##  Changes

  - pyproject.toml: declare the full runtime stack in [project.dependencies], with GPU support (CuPy, kvikio, nvcomp) as hard requirements rather than optional extras. Add [project.optional-dependencies] for dev / docs / moments.
    - Uses cupy-cuda12x[ctk] so the CUDA toolkit headers CuPy needs to JIT-compile its kernels are pulled from PyPI 
    - Adds matplotlib, seaborn (imported by the plotting module via __init__) and tskit (used directly by HaplotypeMatrix)
    - Fixes the Homepage URL to the kr-colab org.
  - pixi.toml: pyproject is now the single source of truth for the PyPI/GPU stack; the duplicated deps are removed and come transitively via the editable install. Dropped the NVIDIA extra-index-url (kvikio-cu12 / nvidia-nvcomp-cu12 are now on the default PyPI index). Added conda cuda-cudart-dev so CuPy's NVRTC compilation finds toolkit headers inside the pixi/conda environment too.
  - Docs: README and docs/source/installation.rst gain an "Installation into an Existing Environment" section; pixi remains the recommended
  path.

 ##  Install paths
```bash
  # Recommended (pinned, reproducible)
  pixi install && pixi shell

  # Into an existing conda/venv environment
  pip install "git+https://github.com/kr-colab/pg_gpu.git"
```

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2026 Andrew Kern
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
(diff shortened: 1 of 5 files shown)


andrewkern:
@mufernando will you try this branch out on your setup?

nspope:
Why are we pinned to python <3.13? This is going to be annoying when integrating pg_gpu into existing pipelines/environments. Is it because of CUDA toolchain or kvikio? Glancing through the lock file, it's not obvious to me where the constraint is.

andrewkern:
ah good catch. we're not in capping the install in `pyproject.toml`. this is vestigal and i'll lift it.

andrewkern:
also -- i'm 95% confident everything (including the cuda stack) installs via pip cleanly, so we can publish to pypi

nspope:
installs and imports fine for me with
```bash
python3 -m pip install "git+https://github.com/kr-colab/pg_gpu.git@feature/pip-installable-deps"
python3 -c "import pg_gpu"
```
(whereas main branch fails as expected on import)

andrewkern:
excellent. when we merge this we can publish to pypi

mufernando (approved):
sorry for the delay! it works for me as well. thanks, @andrewkern!

Issue #121, opened 2026-06-12
Support haploid / hemizygous samples (sex chromosomes); stop silently mishandling them


andrewkern:
## Summary

pg_gpu has no notion of haploid / hemizygous samples. When a VCF (or zarr store) contains haploid or pseudo-diploid calls -- the typical situation for the X/Y in males, the whole genome of haplodiploid or haploid organisms, or organellar contigs -- the loaders silently produce a haplotype matrix with the wrong number of haplotypes and biased statistics, with no warning. This was reported by @jiseonmin in #120 (male-only flies on the X: "the output looks weird since it assumes everyone is homozygous everywhere").

This issue proposes (1) making the failure loud across every loader, and (2) adding first-class opt-in haploid support.

## The problem

A user loads an X-chromosome VCF containing only males. Two encodings are common in the wild, and pg_gpu mishandles both:

- **Pseudo-diploid homozygous** (`GT=1/1`, `0/0`) -- what GATK and many pipelines emit for haploid regions. Every male is read as a homozygous diploid, so each contributes **two identical haplotypes**. The reported `n_haplotypes` is doubled, within-individual LD is perfect, heterozygosity is zero everywhere, and the SFS/pi/Tajima's D are computed over 2N pairwise-identical pseudo-haplotypes -- biased and wrong, but no error.
- **True haploid** (`GT=1`, `GT=0`) -- a single allele per call.

The reporter ran `utils/genome_scan.py`, which routes VCFs straight through `HaplotypeMatrix.from_vcf` (`utils/genome_scan.py:121`).

## Current state of the codebase

**`HaplotypeMatrix.from_vcf`** (`pg_gpu/haplotype_matrix.py:480-487`) is the path the reporter hit:

```python
genotypes = allel.GenotypeArray(vcf['calldata/GT'])
num_variants, num_samples, ploidy = genotypes.shape
assert ploidy == 2

haplotypes = np.empty((num_variants, 2 * num_samples), dtype=genotypes.dtype)
haplotypes[:, :num_samples] = genotypes[:, :, 0]
haplotypes[:, num_samples:] = genotypes[:, :, 1]
```

The `assert ploidy == 2` is effectively dead code, because `allel.read_vcf` defaults to a fixed ploidy-2 layout regardless of the VCF's actual ploidy. I verified both cases against scikit-allel:

| VCF encoding | `calldata/GT` returned | `ploidy` axis | assert |
|---|---|---|---|
[...]

Issue #120, opened 2026-06-12
Silent "failure" on sex chromosomes?


jiseonmin:
Hi colab,

First, thank you so much for creating and sharing an awesome tool! I can confirm it was pretty straightforward to install (using pixi) and run smoothly (and so fast..) on my university cluster . But I wanted to flag one thing about ploidy. I was able to run `genome_scan.py` on X chromosome on a vcf file with just male flies, and it ran fully without a warning. However, the output looks weird since it assumes everyone is homozygous everywhere. Then I saw line 482 in `pg_gpu/haplotype_matrix.py` (`assert ploidy == 2`) inside `from_vcf`, which doesn't seem right, at least without a warning. So my question is, could you edit the line 480 (`genotypes = allel.GenotypeArray(vcf['calldata/GT'])`) and beyond to correct for pseudo-diploids or print warning and stop if there is mixture of haploids and diploids? And I guess if the region includes both autosomes and sex chromosome, it might be a little more work. 
Lastly, I've only used vcf input, so it might be worth checking if similar things happen to other input formats that doesn't depend on `from_vcf`. 

Thanks,
Jiseon

nspope:
Good point---thanks @jiseonmin

andrewkern:
yes thanks @jiseonmin! I'll file an issue to add this as a feature and reference this one.

andrewkern:
Closing this one, with the two things you asked for now in place and the rest tracked in #121.

Stop or warn on haploid and pseudo-diploid data: every loader now runs a diploid-encoding check. A store or VCF whose ploidy axis is not 2, or a sample whose calls are all haploid, raises with the sample named. Data that is polymorphic but never shows a heterozygote, which is what a male-only X looks like after the VCF writer pads it to pseudo-diploid, gets a warning that says so, instead of the silent all-homozygous run you saw. The old `assert ploidy == 2` is gone.

Other input formats: the same check runs in `from_vcf`, in `from_zarr` for both `HaplotypeMatrix` and `GenotypeMatrix`, and on the streaming path, so a zarr store behaves the same way as the VCF.

What remains is real support for haploid and hemizygous samples, so that a male X is analysed as one haplotype per sample rather than rejected or warned about. That is the feature half of #121; please follow it there.

PR #122, opened 2026-06-12
Detect haploid / hemizygous input at load time

andrewkern:
## Summary

Makes pg_gpu's diploid-only loaders fail loudly on haploid and pseudo-diploid
input instead of silently producing doubled haplotype counts and biased
statistics. This is "layer 1" of #121 (the detection/guard layer); first-class
haploid support with an opt-in `ploidy=` / `haploid_samples=` parameter remains
as the follow-on.

Reported by @jiseonmin in #120: running a genome scan on the X chromosome with
a male-only VCF ran without warning but produced output that "assumes everyone
is homozygous everywhere." The `assert ploidy == 2` in `from_vcf` never fired,
because `read_vcf` returns a ploidy-2 layout regardless of the VCF's actual
ploidy.

## What changed

New `check_diploid_encoding(gt, sample_names, source)` (in `pg_gpu/_warnings.py`),
which inspects the actual allele values (not scikit-allel's padded shape) and:

- **Raises `ValueError`** for a non-2 ploidy axis, or for any sample whose calls
  are genuinely haploid -- identified robustly as the `[allele, -1]` padding shape
  with *no* diploid evidence anywhere (a full `[a, b]` call or a reverse `[-1, b]`
  half-call, neither of which haploid padding can produce). This avoids
  misreading a partially-missing diploid call (`GT=1/.`) as haploid.
- **Warns `HaploidDataWarning`** on the unambiguous global tell: a polymorphic
  dataset with zero heterozygous genotypes -- exactly the reporter's "everyone
  homozygous everywhere." A single real heterozygote anywhere suppresses it, so
  it stays quiet on ordinary diploid data where individuals are incidentally
  homozygous in a small region.

Wired into every loader path:

- `HaplotypeMatrix.from_vcf` -- replaces the dead `assert ploidy == 2`
- `HaplotypeMatrix._build_eager` (zarr) -- replaces the manual `gt.shape[2] != 2` check
- `GenotypeMatrix.from_vcf` and `GenotypeMatrix._build_eager` (zarr) -- previously unguarded
- `ZarrGenotypeSource.__init__` (streaming) -- O(1) ploidy-axis guard from store
  metadata, sharing one `_require_diploid_ploidy` helper with the value-based detector

## Warnings consolidation

pg_gpu's warning classes had been accreting one-per-module
(`MemoryLimitedWarning` in `_memory_warning.py`, `BadlyChunkedWarning` in
`streaming_matrix.py`, and now this PR's `HaploidDataWarning`). Rather than add
a third scattered module, this PR consolidates all of them into a single
`pg_gpu/_warnings.py`:

- The three warning classes live together at the top (the taxonomy), followed
  by the load-time helpers that emit them (the VCF size heuristic and the
  diploid-encoding check).
- `_memory_warning.py` and `_ploidy_check.py` are deleted, folded into
  `_warnings.py`; `BadlyChunkedWarning`'s class moves there too (its emit site
  stays in `streaming_matrix`, which now imports the class -- so
  `streaming_matrix.BadlyChunkedWarning` still resolves).
- All three are exported from the package top-level, so every pg_gpu warning is
  silenceable via `from pg_gpu import XxxWarning` /
  `warnings.filterwarnings(category=...)`.

The currently-untyped bare `UserWarning` call sites (in `ld_statistics`,
`windowed_analysis`, `zarr_io`, `zarr_source`) are intentionally left as-is.

## What a user sees now

- Pseudo-diploid male X (`GT=1/1`): loads, with a clear `HaploidDataWarning`
  naming the carriers and the bias it introduces.
- True haploid male X (`GT=1`): `ValueError` naming the samples and explaining
  the half-missing-diploid problem.

## Tests

New `tests/test_ploidy_check.py` (17 tests): unit coverage for normal diploid,
true-haploid (incl. mixed and sample naming), partial-missing and fully-missing
diploids *not* misread as haploid, monomorphic-reference no-warn,
heterozygote-suppresses-warn, ploidy 1/3/4; plus VCF integration tests through
both `from_vcf` loaders. The existing `test_memory_warning.py` repoints to the
consolidated module.

Full suite: 870 passed, 63 skipped. Lint clean.

## Follow-on (not in this PR)

Layer 2 of #121: opt-in `ploidy=` / `haploid_samples=`, mixed-ploidy collapse,
and per-sample handling of pseudo-diploid stores on the streaming path.

Advances #121, addresses #120.


andrewkern:
@nspope this is a PR with the warnings layer

tests/test_ploidy_check.py
```diff
@@ -0,0 +1,171 @@
...
+
+
+@pytest.mark.parametrize("ploidy", [1, 3, 4])
+def test_non_diploid_ploidy_raises(ploidy):
+    gt = np.zeros((4, 3, ploidy), dtype=np.int8)
+    with pytest.raises(ValueError, match=f"ploidy {ploidy}"):
+        check_diploid_encoding(gt, source="test")
+
+
+# --- integration through the VCF loaders -----------------------------------
```
nspope, line 109: any need to test this through zarr loaders too? i.e. dump to vcz and reload?

nspope (approved):
looks good to me ...

tests/test_ploidy_check.py
andrewkern: Good call -- added in 1a22135. The detector is wired into the VCZ `_build_eager` path separately from `from_vcf`, so it warrants its own coverage. New zarr roundtrip tests:

- pseudo-diploid `HaplotypeMatrix` dumped to VCZ via `to_zarr` and reloaded through both `HaplotypeMatrix.from_zarr` and `GenotypeMatrix.from_zarr` -> `HaploidDataWarning`
- a normal diploid (heterozygous) store reloads quietly
- a genuinely haploid store (ploidy-1 `call_genotype` axis) is rejected on load

PR #127, opened 2026-07-09
Multiallelic handling for scalar statistics and SFS

nspope:
First of several staggered PRs onto `nsp-multiallelic-trunk`. This still needs some cleanup, but is ready for feedback (esp on behavior changes and arbitrary decisions).

**Background:** pg_gpu collapsed multiallelic sites so pi, theta estimators, Tajima's D, and the whole SFS family were wrong whenever a site had >2 alleles. This PR adds a per-allele allele-count primitive (`_memutil.allele_counts`) and routes the `diversity` and `sfs` modules through it, matching tskit. The goal is that statistics on biallelic data should be identical, only multiallelic behaviour should change, and that our target for parity is tskit.

**Scope:** `diversity.py` + `sfs.py` (every SFS variant: unfolded / folded / scaled, joint/2D, and hypergeometric projection), plus removal of the duplicate `diversity.allele_frequency_spectrum`. I didn't touch `divergence`, `admixture`, `windowed_analysis`, and `decomposition`; these still collapse multiallelic sites and fixes will follow in later PRs. Also, I've only touched the library itself, auxiliary scripts and documentation still need to be updated.

**Verification:** `tests/test_allele_counts.py` + `tests/test_multiallelic.py` check pi / S / Watterson / Tajima's D / theta_H / theta_L / He / singletons and the full SFS family against tskit and scikit-allel (preference to the former); with a biallelic control and reference-absent, tri-/tetra-allelic, and missing-data edge cases.

## Behaviour changes

Multiallelic stats match tskit: pi, theta_W, theta_H, theta_L, Tajima's D, Fay-Wu H, Zeng E, segregating sites, singletons, He, and every SFS (`sfs`, `sfs_folded`, `*_scaled`, `joint_sfs`, `joint_sfs_folded`, `project_joint_sfs`) are now per-allele, matching tskit. Mutation-count for segregating / theta_W / Tajima's D uses tskit's `(alleles−1)` weighting, which diverges from scikit-allel's polymorphic-site count on multiallelic sites

But there are some conventions that also touch biallelic data (causing a behavior change to earlier pg_gpu):
- SFS fixed classes excluded to match tskit, which changes the SFS / `daf_histogram` on biallelic data only when
  monomorphic/fixed sites are present. All theta values are unchanged.
- Folded SFS dtype to float64 because of tskit's 0.5-weighting for folding
- `joint_sfs_folded` now uses tskit's total-fold, which diverges from scikit-allel even on biallelic data. allel folds each axis independently (which can mix two different alleles into one cell); tskit folds the site by the global minor allele. Output shape also changes `(n1//2+1, n2//2+1)` to `(n1+1, n2+1)`. This is the one place we knowingly change biallelic output beyond the endpoint convention.

## Arbitrary decisions 

These had no tskit-parity guide:
- `FrequencySpectrum` pi / Tajima's D / Fay-Wu H are SFS-based approximations for multiallelic data (because a marginal SFS can't reproduce per-allele pi); the scalar functions (outside FrequencySpectrum) are authoritative and match tskit.
- `max_daf` / `daf_histogram` = per-derived-allele; `mu_sfs` = per-derived-allele edge/segregating fraction. The alternative is to collapse all non-ref alleles into one count, but that'd be inconsistent with the tskit-style handling.
- `joint_sfs_folded` total-fold discussed above: tskit keeps one physical allele per site, allel's per-axis fold can mix alleles).

## Misc

- Removed API for `diversity.allele_frequency_spectrum` and `HaplotypeMatrix.allele_frequency_spectrum`, as these are superceded by the `sfs` module
- `_stream_sum` now preserves the kernel dtype.
- hypergeometric projection works with tskit's multiallelic SFS, even with folding: that is, we'd get the same thing if we did hypergeometric sampling on the "multiallelic spectrum", and then collapsed down to "non-reference".

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -111,7 +111,6 @@ Core Functions
 .. autofunction:: pg_gpu.diversity.theta_w
 .. autofunction:: pg_gpu.diversity.tajimas_d
 .. autofunction:: pg_gpu.diversity.haplotype_diversity
-.. autofunction:: pg_gpu.diversity.allele_frequency_spectrum
 .. autofunction:: pg_gpu.diversity.segregating_sites
 .. autofunction:: pg_gpu.diversity.singleton_count
 .. autofunction:: pg_gpu.diversity.fay_wus_h
--- a/pg_gpu/_memutil.py
+++ b/pg_gpu/_memutil.py
@@ -2,6 +2,9 @@
 
 import cupy as cp
 
+# Threads per block for the one-thread-per-variant fused kernels.
+_THREADS_PER_BLOCK = 256
+
 
 def estimate_variant_chunk_size(n_hap, bytes_per_element=4, n_intermediates=3,
                                  memory_fraction=0.4):
@@ -149,7 +152,7 @@ def dac_and_n(hap):
     out_dac = cp.empty(n_var, dtype=cp.int64)
     out_n = cp.empty(n_var, dtype=cp.int64)
     s0, s1 = hap.strides
-    threads = 256
+    threads = _THREADS_PER_BLOCK
     blocks = (n_var + threads - 1) // threads
     _dac_and_n_kernel((blocks,), (threads,),
                       (hap, n_hap, n_var, s0, s1, out_dac, out_n))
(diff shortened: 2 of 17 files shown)


andrewkern:
looks great @nspope. I pulled it locally and messed about. You note this above but I think we have to update specifically the `quickstart.rst` and the `features.rst`.  The arbitrary decisions here are fine, and I like that the folded will mimic tskit. 

I'm going to merge it, so that you can get the follow up PRs in next.

andrewkern (approved):

PR #128, opened 2026-07-10
Multiallelic handling in count-based divergence stats

nspope:
**Background:** The count-based two-population statistics in `divergence.py` collapsed multiallelic sites, so dxy, the Fst estimators, and PBS were wrong at any site with more than two alleles. This PR routes them through the per-allele `_memutil.allele_counts` primitive and adds `fst_tskit` for tskit parity. Biallelic values are unchanged; only multiallelic behaviour changes, plus the one Weir-Cockerham divergence noted below.

**Scope:** `dxy`, `dxy_components`, `da`, `fst_hudson`, `fst_tskit` (new), `fst_nei`, `fst_weir_cockerham`, `pbs`, and the aggregators (`fst`, `divergence_stats`, `pairwise_fst`). Out of scope (to be dealt with later): the distance-based stats (`snn`, `dxy_min`, `gmin`, `dd`, `dd_rank`, `zx`) and `distance_stats._pairwise_diffs_matrix_gpu`, which is separately multiallelic-broken (its `sum_i + sum_j - 2*X@X.T` assumes binary genotypes and returns negative distances on multiallelic pairs). Also, doc/auxiliary script updates are deferred. The `windowed_analysis` analogue is left uncorrected, and will be addressed later.

**Behaviour changes:** Stats corrected for multiallelic data (these were collapsed before):

- `dxy` / `dxy_components` / `da` are now per-allele `1 - sum_a p1_a*p2_a`, matching tskit `divergence`.
- `fst_hudson` / `pbs` use per-allele within/between diversity with the classic assembly `1 - Hw/Hb`, matching `allel.hudson_fst` / `allel.pbs`.
- `fst_nei` uses per-allele `HS/HT` with within-heterozygosity `het = 1 - sum_a p_a^2`.
- `fst_weir_cockerham` uses per-allele one-vs-rest `a/b/c` components summed over alleles, matching `allel.weir_cockerham_fst`.

New function: `fst_tskit` computes `(sum(Hb) - sum(Hw)) / (sum(Hb) + sum(Hw))`, tskit's combination, which is a distinct assembly from classic Hudson (`fst_hudson` is unchanged and still the classic `1 - Hw/Hb`). It is wired into `fst(method='tskit')` and `divergence_stats`.

The one genuine divergence: `fst_weir_cockerham`'s c component now uses per-allele observed heterozygosity (individuals carrying exactly one copy of allele a), not the site-level "any difference" heterozygosity the old biallelic code used. This is required because WC is a per-allele ANOVA and c is the between-gamete-within-individual variance of that specific allele; the two coincide for biallelic sites, so biallelic output is unchanged. The public `diversity.heterozygosity_observed` is deliberately left as the standard site-level H_obs (it already matches `allel.heterozygosity_observed`, is bounded in [0,1], and is perfectly correct for multiallelic data).

**Verification:** New `tests/test_multiallelic.py::TestDivergenceVsTskit` covers: dxy and da vs tskit `divergence`; `fst_hudson`, `pbs`, and `fst_weir_cockerham` vs scikit-allel; `fst_tskit` vs `ts.Fst`; and `fst_nei` vs the per-allele hand formula (scikit-allel has no Nei GST).

### DIFF
--- a/pg_gpu/divergence.py
+++ b/pg_gpu/divergence.py
@@ -243,6 +243,40 @@ def fst_tskit(haplotype_matrix: HaplotypeMatrix,
     return 0.0
 
 
+def _pop_wc_stats(pop_haps, k):
+    """Per-allele diploid stats for Weir & Cockerham from one population.
+
+    Consecutive haplotypes are one diploid individual (a trailing unpaired
+    haplotype is dropped). For each allele ``a`` in ``0..k-1`` returns the
+    number of copies and the number of individuals heterozygous FOR THAT ALLELE
+    (exactly one copy) -- the per-allele observed het that WC's c component
+    needs (see the NOTE in ``fst_weir_cockerham``).
+
+    Returns
+    -------
+    ac : cupy.ndarray, float64, shape (n_variants, k)
+        Allele copies from complete diploid pairs.
+    het : cupy.ndarray, float64, shape (n_variants, k)
+        Per-allele heterozygote counts.
+    n : cupy.ndarray, float64, shape (n_variants,)
+        Complete diploid individuals per site.
+    """
+    n_pairs = pop_haps.shape[0] // 2
+    ha = pop_haps[0:2 * n_pairs:2, :]
+    hb = pop_haps[1:2 * n_pairs:2, :]
+    both = (ha >= 0) & (hb >= 0)
+    n = cp.sum(both, axis=0).astype(cp.float64)     # complete individuals/site
+    n_var = pop_haps.shape[1]
+    ac = cp.zeros((n_var, k), dtype=cp.float64)      # allele copies (from pairs)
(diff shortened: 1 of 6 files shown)


andrewkern (approved):
One thing I see here, is the same class as the pi issue in #127: scalar dxy/fst are corrected, but the windowed_analysis fused/scatter kernels still collapse alt alleles into one derived count, so windowed dxy/fst/da/fst_wc diverge from the corrected scalars on multiallelic data. #127 set up a strict `xfail` for pi but here there is no matching test. Might be worth an analogous xfail or just adding the windowed FST/dxy kernels to the out-of-scope list so it's tracked.

nspope:
Good point, I was hoping to do the windowed analysis in one pass after covering all the other stats. But I'll add the xfail, so we're tracking it.

PR #129, opened 2026-07-10
Correct pairwise distances for multiallelic data

nspope:
**Background:** the haploid pairwise-distance kernel `distance_stats._pairwise_diffs_matrix_gpu` computed Hamming distance via the binary-only identity `diffs_ij = sum_i + sum_j - 2*(X@X.T)_ij` on the raw allele-index matrix, which only holds for `{0,1}` data (it silently assumes `v^2 = v`). On multiallelic allele indices it returned garbage (negative "distances"). This PR replaces it with the one-hot form `diffs = joint_valid - sum_v (I_v @ I_v.T)` (count same-allele sites over allele values, subtract from the jointly-valid count), which is multiallelic-correct and reduces to the old values exactly on biallelic no-missing data. Every distance-based two-population statistic inherits the fix through the kernel, so no consumer code changed.

**Scope:** the kernel `_pairwise_diffs_matrix_gpu`, `pairwise_diffs_haploid`, and the divergence distance stats via `pairwise_distance_matrix` (`snn`, `dxy_min`, `gmin`, `dd`, `dd_rank`, `distance_based_stats`). Out of scope by deliberate decision: `pairwise_diffs_diploid` (see the decision below). Also, docs and auxiliary scripts are untouched and will be updated later.

### Behaviour changes

Bugfix: missing data in the kernel's `'include'` mode (also affects biallelic data). The old code set missing genotypes to allele 0 (`x = where(hap>=0, hap, 0)`) and never masked them, so a missing call was treated as a reference homozygote: a (missing, allele-1) pair counted as a difference, a (missing, missing) pair as a match, and missing sites sat in the implicit denominator. The one-hot rewrite excludes missing entries from both the match count and `joint_valid`, so distances are over jointly-non-missing sites only (what the docstring erroneously promised).

Multiallelic correction: pairwise Hamming distances (and every stat built on them) are now correct for allele indices beyond `{0,1}`, where the old kernel produced meaningless (often negative) values.

Out-of-scope: `pairwise_diffs_diploid` is left biallelic-only. We could respec it for multiallelic, but chose not to: its only input, `GenotypeMatrix`, is biallelic by construction (0/1/2 alt dosage; `from_vcf` filters multiallelic sites; `from_haplotype_matrix` sums paired haplotypes, defined only for `{0,1}`), so supporting multiallelic needs a new genotype representation plus a one-hot-over-`K(K+1)/2`-genotypes kernel (quadratic in K). This seems unnecessary, because multiallelic diploid distance is already available through the haploid allele-index path. The biallelic-only contract is now made explicit in the `pairwise_diffs_diploid` and `GenotypeMatrix` docstrings.

Cost: the kernel does K matmuls per variant-chunk (was 1), linear in the number of allele values K.

### DIFF
--- a/pg_gpu/distance_stats.py
+++ b/pg_gpu/distance_stats.py
@@ -32,6 +32,15 @@ def _pairwise_diffs_matrix_gpu(hap, missing_data='include'):
     chunks to GPU on-the-fly so the full matrix never needs to reside
     on GPU at once.
 
+    Multiallelic-correct Hamming: the number of DIFFERING jointly-valid sites is
+    ``joint_valid - same``, where ``same`` counts sites at which the two
+    haplotypes carry the same allele, summed over allele values via one-hot
+    indicators (``same = sum_v I_v @ I_v.T``, ``I_v[h,s] = 1 if hap[h,s] == v``).
+    This generalizes the biallelic ``sum_i + sum_j - 2*X@X.T`` gram identity
+    (which silently assumes ``v^2 = v``) to any number of alleles, and reduces
+    to it exactly on ``{0, 1}`` data. Missing entries (``-1``) match no allele
+    value and are excluded from ``joint_valid``, so they never contribute.
+
     Parameters
     ----------
     hap : numpy.ndarray or cupy.ndarray, shape (n_haplotypes, n_variants)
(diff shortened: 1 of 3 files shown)


andrewkern (approved):
this one looks GTG

PR #130, opened 2026-07-10
Multiallelic admixture statistics

nspope:
**Background:** `admixture.py`'s f-statistics built a single "alt frequency" via `_allele_freq = cp.sum(hap)/n`, which for allele-index haplotypes sums allele INDICES (so a frequency can exceed 1). So f2/f3/f4/D were wrong on multiallelic sites. 

**Scope:** This PR routes f2/f3/f4 through per-allele counts matching tskit, adds a public `patterson_f4`, and makes `patterson_d` biallelic-restricted. Biallelic data is unchanged. The moving admixture statistics inherit the fix as they utilize the same underlying machinery. Docs/auxiliary scripts are not updated, that's deferred til later.

### What changed

f2 and f3 (`patterson_f2`, `patterson_f3`) now use tskit's unbiased per-allele U-statistic (the exact `f2_summary_func` / `f3_summary_func` from the tskit C source), summed over alleles. They match `TreeSequence.f2` / `TreeSequence.f3` on multiallelic data and are unchanged on biallelic data (the ADMIXTOOLS `(a-b)^2 - het` form already equals tskit's all-alleles sum there). f3 still returns `(T, B)`, now with `T` = tskit f3(C;A,B) and `B` = the unbiased heterozygosity of the target C (which equals `ts.diversity(C)`).

New `patterson_f4(A,B,C,D)` -- the per-allele unbiased f4, matching `TreeSequence.f4`. tskit exposes f4 and pg_gpu previously had no public f4 (it only existed implicitly inside D), so this is a parity addition.

`patterson_d` (ABBA-BABA) is now biallelic-restricted (it silently filters to biallelic). D is a biallelic-SNP statistic and no library (tskit or scikit-allel) computes a multiallelic D as far as I know, although f4 is proportional to the numerator of D so a generalization is certainly possible (but deferred here). The retained sites use proper per-allele counts (the frequency of the single alt allele, alt = highest-index present allele), so even a biallelic `{0,2}`-type site is correct rather than index-inflated; standard `{0,1}` data is unchanged. Exclusion is silent, consistent with the package's other biallelic filtering (e.g. `GenotypeMatrix.from_vcf` drops multiallelic sites without warning).

Removed the dead `_allele_freq` / `_allele_freq_and_het` helpers (the index-inflation source; unreferenced after the rewrite).

### DIFF
--- a/pg_gpu/admixture.py
+++ b/pg_gpu/admixture.py
@@ -13,49 +13,76 @@ from ._utils import get_population_matrix as _get_population_matrix
 from .resampling import block_jackknife, _moving_nansum, _moving_nanmean
 
 
-def _allele_freq(haplotype_matrix):
-    """Compute alternate allele frequency from per-site valid data."""
-    if haplotype_matrix.device == 'CPU':
-        haplotype_matrix.transfer_to_gpu()
+def _aligned_allele_counts(haplotype_matrix, pops):
+    """Per-allele counts for a list of populations on a shared global K.
 
-    hap = haplotype_matrix.haplotypes
-    valid_mask = hap >= 0
-    n_valid = cp.sum(valid_mask, axis=0).astype(cp.float64)
-    n1 = cp.sum(cp.where(valid_mask, hap, 0), axis=0).astype(cp.float64)
-    return cp.where(n_valid > 0, n1 / n_valid, 0.0)
-
-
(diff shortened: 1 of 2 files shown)


andrewkern (approved):
LGTM Nate

PR #132, opened 2026-07-11
Fix windowed analysis for multiallelic data (with some exceptions)

nspope:
Based on #130 for a clean diff.

**Background:** `windowed_analysis.py` computes its scalar statistics through three separate engines -- a host scatter-add path, fused CUDA kernels, and a per-variant path. All three collapsed multiallelic sites, so windowed pi/theta/Tajima/segregating/singletons and dxy/fst/da were wrong on multiallelic sites. This PR makes every windowed engine per-allele so each window equals the scalar function on that window's variants (with some exceptions that require discussion). Biallelic data is unchanged except where noted below.

### What changed

**Single-pop, all three engines**:
- Host scatter `_windowed_thetas_scatter`: `_prepare_dac`/`_site_contribution` changed to `diversity._prepare_allele_counts`/`_ac_contribution`. Covers pi, theta_w/h/l, tajimas_d, segregating_sites, singletons, the Fay-Wu/Zeng neutrality tests, max_daf.
- Fused kernel `_fused_windowed_kernel_v2` (drives `windowed_statistics_fused` and `_chunked`): uses per-variant `cnt[MAX_ALLELES]` array instead of a single `dac`. Covers per-allele pi / mutation-count segregating / singleton / theta_h / max_daf.
- Per-variant `windowed_statistics`: per-allele pi, theta_w, segregating_sites, singletons, tajimas_d, and "het_expected" (which only this engine exposes).

**Two-pop, all three engines**:
- `_twopop_site_components` (used by both the scatter engine and the per-variant path) and `_fused_windowed_twopop_kernel`: per-allele within-pop (`sum_a c(c-1)`) and between-pop (`sum_a c1*c2`) pair counts 
- The fused kernel's `fst_wc` output is left untouched (see Deferred).

**Scalar bugfix (divergence.py):** `dxy(span_normalize=False)` (and therefore `da(span_normalize=False)`) returned `dxy_sum / n_sites` while `pi` and every other statistic return the raw sum. Now `dxy` falls through to the
shared `_apply_span_normalize` and returns the raw sum.

**MAX_ALLELES cap:** the fused CUDA kernels need a fixed-size per-allele array with hardcoded size `MAX_ALLELES = 8`. Sites with an allele index >= 8 are dropped before kernel launch with a new `MultiallelicCapWarning`. The added overhead seems to be ~2% of the end-to-end call. However the `cnt[]` array stays in registers (out of local memory) so it is sample-size independent. However we should benchmark further when this whole multiallelic thrust is wrapped up.

**Cleanup:** removed the now-dead collapsed helpers `_allele_sum_and_n`, `_per_variant_mpd`, `_per_variant_is_seg`, `_per_variant_is_singleton`.

## Behaviour changes

We want the windowed statistics to mirror the scalar path---we should always try to be internally consistent.

**Multiallelic corrected**: pi, theta_w/h/l, tajimas_d, segregating_sites, singletons, max_daf, het_expected, dxy, fst/fst_hudson, da from all windowed engines now match the per-allele scalar functions per window if there's no missing data (see below). Output on biallelic data is unaffected.

**Span normalisation bugfix in `divergence.dxy` / `da`:** see description above.

**`windowed_statistics` changes for consistency with scalar path:**
- `singletons`: was folded (`dac==1 or dac==n-1`, counting ancestral-allele singletons too); now unfolded (derived alleles observed exactly once.
- `segregating_sites`: now the mutation count (`#alleles_present - 1`); identical to the old boolean is-segregating flag for biallelic and matches tskit for multiallelic.
- `theta_w` / `tajimas_d` numerator: now use the per-site `a1(n_valid)` instead of a single `a1(n_hap)`. Identical without missing data; corrected under missing data.

## Deferred (tracked as separate issues)

1. Windowed `fst_wc` (Weir-Cockerham) computed by the fused kernel is a haploid estimator, and disagrees with the scalar diploid `fst_weir_cockerham` even on biallelic data. This PR didn't touch fst_wc, as it will likely need to be rewritten for parity.

2. Windowed neutrality-test variance uses nominal `n_hap`, while the scalar uses the harmonic-mean effective `n` over per-site `n_valid`. So windowed `tajimas_d` (and `normalized_fay_wu_h`, `zeng_e`, `zeng_dh`) diverge from the scalar under missing data.

3. We need some more stringent testing for cross-implementation parity

## Verification

New coverage in `tests/test_windowed_analysis.py`:
- `TestMultiallelicSinglePop`: scatter, fused, and per-variant engines vs scalar for the six single-pop stats on a genuinely multiallelic (JC69) fixture; plus a missing-data parity test (pi/theta_w/segregating_sites/singletons/het_expected pass) and a `xfail(strict)` missing-data `tajimas_d` test (the effective-n issue above).
- `TestMultiallelicTwoPop`: scatter, fused, and per-variant engines vs scalar for dxy/fst_hudson/da on the multiallelic fixture.

### DIFF
--- a/pg_gpu/__init__.py
+++ b/pg_gpu/__init__.py
@@ -65,9 +65,10 @@ from .decomposition import (
 from .resampling import block_jackknife, block_bootstrap
 from ._warnings import (
     BadlyChunkedWarning, HaploidDataWarning, MemoryLimitedWarning,
+    MultiallelicCapWarning,
 )
 
-__all__ = ['ld_statistics', 'diversity', 'divergence', 'windowed_analysis', 'selection', 'sfs', 'admixture', 'decomposition', 'plotting', 'distance_stats', 'resampling', 'HaplotypeMatrix', 'GenotypeMatrix', 'WindowedAnalyzer', 'windowed_analysis', 'AccessibleMask', 'bed_to_mask', 'parse_bed', 'LocalPCAResult', 'LostructResult', 'local_pca', 'local_pca_jackknife', 'lostruct', 'pc_dist', 'corners', 'block_jackknife', 'block_bootstrap', 'MemoryLimitedWarning', 'HaploidDataWarning', 'BadlyChunkedWarning']
+__all__ = ['ld_statistics', 'diversity', 'divergence', 'windowed_analysis', 'selection', 'sfs', 'admixture', 'decomposition', 'plotting', 'distance_stats', 'resampling', 'HaplotypeMatrix', 'GenotypeMatrix', 'WindowedAnalyzer', 'windowed_analysis', 'AccessibleMask', 'bed_to_mask', 'parse_bed', 'LocalPCAResult', 'LostructResult', 'local_pca', 'local_pca_jackknife', 'lostruct', 'pc_dist', 'corners', 'block_jackknife', 'block_bootstrap', 'MemoryLimitedWarning', 'HaploidDataWarning', 'BadlyChunkedWarning', 'MultiallelicCapWarning']
 
 # Version is derived from the git tag at build time (hatch-vcs) and read here
 # from the installed package metadata, so there is no hardcoded string to keep
--- a/pg_gpu/_warnings.py
+++ b/pg_gpu/_warnings.py
@@ -67,6 +67,23 @@ class BadlyChunkedWarning(UserWarning):
     unlock the GPU-decode fast path."""
 
 
+class MultiallelicCapWarning(UserWarning):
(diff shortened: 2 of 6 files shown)


andrewkern:
this look great Nate. I've got one request, which is that you add a test to trip the MAX ALLELES cap

nspope:
ok should be good (and it has the right base for the merge now). Merge whenever!

nspope:
I'm going to go ahead and merge this, to keep things moving along

Issue #135, opened 2026-07-11
Pin statistics across different implementations in test suite


nspope:
There are four distinct paths to compute the same statistics (three engines in windowed_analysis and the scalar computations). Sometimes five, given the FrequencySpectrum class. There's nothing in the test suite asserting equivalence the distinct implementations across missing data / multiallelic data / etc. We should have a testing suite that is responsible for comparing them, with explicit xfail for exceptions (though ideally there wouldn't be exceptions).

andrewkern:
i'm going to work on addressing this today

nspope:
sounds good, let's get #132 in first as that should make a number of them match, i'll push the missing test & merge shortly

nspope:
mostly closed, followups in #146

PR #137, opened 2026-07-13
Pin statistics across implementations in a parity test suite

andrewkern:
## Summary

Addresses #135. Adds `tests/test_implementation_parity.py`, a suite that asserts the same population-genetic statistic agrees across the code paths that compute it: scalar functions, `FrequencySpectrum`, and the windowed **scatter**, **fused** CUDA, and **python-loop** engines.

Targets `nsp-multiallelic-trunk` (this suite validates the multiallelic work), not `main`.

Each statistic is described once in a declarative `_STATS` registry; the parametrized test drives every supported path across four data conditions — clean, missing/include, missing/exclude, and multiallelic — and asserts it matches the scalar reference. **16 statistics, ~180 cases: 149 passed, 12 skipped, 19 xfailed.** GPU-only (skipped without CUDA).

This is pure test infrastructure — no library code changes. Divergences between paths are recorded as `strict=True` xfails in `_XFAILS` with a mechanism-level reason; a `(stat, path)` may carry several condition-scoped rules. Fixing any divergence removes its rule and flips the cell to a pass.

## Divergences this suite currently pins

1. **`fst_wc` (fused; all conditions)** — the fused kernel was not converted to the per-allele one-vs-rest Weir-Cockerham ANOVA the scalar/python-loop path now uses.
2. **`theta_w`, `tajimas_d` (fused; missing/include)** — the fused kernel's per-site sample-size handling under missing data still diverges from the scalar.
3. **`tajimas_d`, `normalized_fay_wu_h`, `zeng_e` (fs and scatter; missing/include)** — the Achaz neutrality-test variance uses a modal / full-sample size rather than per-site valid counts.
4. **`pi`, `theta_w`, `tajimas_d`, `segregating_sites`, `fay_wu_h`, `normalized_fay_wu_h`, `zeng_e` (fs; multiallelic)** — `FrequencySpectrum` builds a per-allele SFS whose segregating-class count and pi differ from the scalar's per-allele counts.
5. **`da` (scatter; missing/exclude)** — the within-population pi terms use a different site set than the scalar (`dxy` agrees).

## Notes

- The fused engine only runs under `missing_data='include'`, so it is skipped (not failed) under the exclude condition.
- Two-population fixtures use an even haplotype count per population, which `fst_weir_cockerham` requires for its diploid pairing.

### DIFF
--- a/tests/test_implementation_parity.py
+++ b/tests/test_implementation_parity.py
@@ -1,0 +1,463 @@
+"""Cross-implementation parity tests.
+
+The same population-genetic statistics are computed by several independent code
+paths in pg_gpu:
+
+  1. scalar functions            -- ``diversity.pi``, ``divergence.dxy``, ...
+  2. ``FrequencySpectrum``       -- SFS dot-product path
+  3. windowed "scatter" engine   -- ``_windowed_thetas_scatter`` / ``_windowed_twopop_scatter``
+  4. windowed "fused" engine     -- ``windowed_statistics_fused`` (CUDA kernels)
+  5. windowed "python-loop"      -- ``WindowedAnalyzer`` per-window fallback
+
+Nothing else in the suite asserts these agree. This module does. Each statistic
+is described once in the ``_STATS`` registry -- which paths compute it, how to
+invoke each, and the span-independent "canonical quantity" to compare -- and the
+parametrized test drives every supported path across four data conditions
+(clean, missing/include, missing/exclude, multiallelic) and asserts it matches
+the scalar reference.
+
+Every path is invoked so its return value is the same span-independent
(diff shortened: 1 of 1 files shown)


nspope:
did you base off of #132 [that is, merge target is nsp-multiallelic-trunk]? I think that fixed the singleton and da and fst issues, at least

andrewkern:
shit, I did not. not me pull in those changes to this branch

nspope:
i'd just rebase on nsp-multiallelic-trunk, that's where i'm doing all the multiallelic stuff so I don't break main in the meantime

andrewkern:
okay this should be ready to go now @nspope

PR #143, opened 2026-07-14
Make the PCA standardization multiallelic-correct.

nspope:
EDIT: updated with the new tskit-analogous verison

decomposition's PCA standardization was wrong on multiallelic input: it derived per-site centering and scaling from a collapsed derived-allele count and centered the raw allele-index matrix, so a site with an allele index >= 2 fed the arbitrary integer allele label into the decomposition. Principal components, explained variance, and every local PCA window covariance became a function of arbitrary allele numbering, silently and non-reproducibly under relabeling.

Rather than patch the single standardization, this splits the family along the data type, because there are two established and genuinely different PCAs and they belong to different inputs:

- pca / randomized_pca take a HaplotypeMatrix and compute the multi-allele haploid PCA: an eigendecomposition of the genetic relatedness matrix. Every present allele (including the reference) is one-hot encoded, each column centered by its allele frequency with no variance scaling, and the sample-by-sample Gram is normalized by the number of segregating sites. Each haplotype is its own sample, so this is multiallelic-correct at any ploidy.
- pca_dosage / randomized_pca_dosage take a GenotypeMatrix and compute the diploid dosage PCA standardized by binomial variance: each biallelic variant's alt-allele dosage (0/1/2) is centered on its mean and scaled by the binomial standard deviation sqrt(p (1 - p)), p = mean / 2, then the individual-by-individual Gram is eigendecomposed. This is the classical dosage PCA and reproduces scikit-allel's diploid PCA exactly.

Each entry point rejects the other matrix type with a TypeError that names its counterpart. local_pca, lostruct, and the local-PCA jackknife move to the same all-allele centered standardization as pca (they remain HaplotypeMatrix-only). The scaler menu (patterson / standard / None) is dropped in favor of the two fixed conventions.

#### What changed

pca and randomized_pca now build the all-allele centered one-hot (all present alleles, centered by frequency, no scaling, missing imputed to the frequency so it contributes zero) and normalize the Gram by the segregating-site count. That Gram is the genetic relatedness matrix, so pca is pinned against a tree sequence's genetic_relatedness output (all sample pairs, centered, proportion-normalized).

pca_dosage and randomized_pca_dosage are new. They standardize the biallelic diploid dosage by centering and binomial-variance scaling, then eigendecompose the individual Gram with no segregating-site normalization, so the coordinates match scikit-allel's magnitude as well as its subspace. They reject dosages > 2 (which cannot arise from a biallelic diploid site) and reject a HaplotypeMatrix. Population subsetting is done inline over individuals; the shared population helper is haplotype granularity on this branch, and unifying genotype/haplotype subsetting is a logged follow-up.

local_pca, lostruct, and local_pca_jackknife route each window through the all-allele centered standardization instead of the old per-variant standardization; they keep their own per-window covariance X @ X.T / (ncol - 1). Adjacent windows' per-PC eigenvectors are sign-aligned (flip when the dot with the previous valid window is negative) so windowed PC tracks are continuous.

The obsolete standardization machinery is deleted: the split biallelic/multiallelic Patterson-scaled helpers and the memory-constrained deferred-PCA path. The four public functions are exported from the package.

#### Behaviour

pca and randomized_pca change on all input, not just multiallelic: the standardization is now all-allele centered with no variance scaling and a segregating-site denominator, where before it was a per-variant Patterson standardization on haplotypes. Explained-variance ratios and the top subspace change accordingly; the output is the eigendecomposition of the genetic relatedness matrix.

The diploid dosage PCA (the previous default's biallelic behaviour) is still available, now as pca_dosage on a GenotypeMatrix.

local_pca and lostruct change on all input for the same reason, and no longer bit-match the published R local-PCA on biallelic data: the all-allele one-hot keeps a column per present allele where R keeps one dosage column per site, and it is single-centered (variant only) where R's classical-MDS double-centers (variant and sample). The two agree for genealogical structure but diverge for structure that aligns with average distance-from-reference (a per-group frequency shift); the reasoning is in the centering rationale note. The strict R-parity tests are xfailed with that reason.

#### Verification

The all-allele Gram equals a tree sequence's genetic_relatedness matrix, and pca's explained variance equals the eigenvalues of that matrix. randomized_pca matches full pca (explained variance plus top subspace). pca_dosage reproduces scikit-allel's diploid PCA explained-variance ratios to 1e-8 and coordinates up to a per-component sign; randomized_pca_dosage matches pca_dosage on structured data via canonical correlations. Type guards and the biallelic requirement are covered. Per-window local_pca / lostruct / jackknife match a direct per-window computation, adjacent windows are sign-aligned, and the R-parity suite is xfailed where the conventions genuinely differ.

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -264,6 +264,8 @@ PCA
 
 .. autofunction:: pg_gpu.decomposition.pca
 .. autofunction:: pg_gpu.decomposition.randomized_pca
+.. autofunction:: pg_gpu.decomposition.pca_dosage
+.. autofunction:: pg_gpu.decomposition.randomized_pca_dosage
 
 Distance
 ~~~~~~~~
--- a/docs/source/examples.rst
+++ b/docs/source/examples.rst
@@ -190,8 +190,7 @@ PCA and Dimensionality Reduction
    from pg_gpu import decomposition
 
    # GPU-accelerated PCA (up to 56x faster than allel)
-   coords, var_ratio = decomposition.pca(h, n_components=10,
-                                          scaler='patterson')
+   coords, var_ratio = decomposition.pca(h, n_components=10)
 
    # Randomized PCA for very large datasets
    coords, var_ratio = decomposition.randomized_pca(
--- a/docs/source/examples.rst
+++ b/docs/source/examples.rst
@@ -189,16 +189,17 @@ PCA and Dimensionality Reduction
 
    from pg_gpu import decomposition
 
-   # tskit PCA: eigendecomposition of the genetic_relatedness matrix
-   # (all alleles kept, multiallelic-correct). Takes a HaplotypeMatrix.
+   # Multi-allele haploid PCA: eigendecomposition of the genetic relatedness
+   # matrix, keeping all alleles (multiallelic-correct). Takes a HaplotypeMatrix.
    coords, var_ratio = decomposition.pca(h, n_components=10)
 
    # Randomized approximation for very large datasets
    coords, var_ratio = decomposition.randomized_pca(
        h, n_components=10, random_state=42)
 
-   # Patterson/GCTA PCA of biallelic diploid dosages (EIGENSTRAT, matches
-   # scikit-allel). Takes a GenotypeMatrix; has a randomized counterpart.
+   # Diploid dosage PCA standardized by binomial variance (center each biallelic
+   # variant's dosage, scale by sqrt(p(1-p))). Takes a GenotypeMatrix; has a
+   # randomized counterpart.
    from pg_gpu import GenotypeMatrix
    g = GenotypeMatrix.from_haplotype_matrix(h)  # pair haplotypes as diploids
    coords, var_ratio = decomposition.pca_dosage(g, n_components=10)
--- a/docs/source/examples.rst
+++ b/docs/source/examples.rst
@@ -189,12 +189,21 @@ PCA and Dimensionality Reduction
 
    from pg_gpu import decomposition
 
-   # GPU-accelerated PCA (up to 56x faster than allel)
+   # tskit PCA: eigendecomposition of the genetic_relatedness matrix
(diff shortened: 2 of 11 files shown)


nspope:
Looking at this again, I'm not sure matching scikit-allel is the right choice---that is, summing over all non-reference alleles. tskit doesn't have site-PCA implemented yet, but for the GRM it sums over all alleles including the reference. That'd be the more internally consistent option here (... this is unlikely to matter in practice, but might as well make the semantics all internally consistent)

andrewkern:
i think we should generally aim to match tskit behavior over scikit-allel here

nspope:
Sounds good. So I think this also means returning an unstandardized (but optionally centered) GRM for the relatedness issue ... unlike GCTA, which uses a 2p(1-p) standardization. If/when we decide on a multiallelic standardization we can add a `standardized=True` option.

andrewkern:
sounds good to me

nspope:
so downside here: if we make everything consistent, then lostruct will differ from the R implementation (which "double-centers", as it centers each sample as well as each site). That's an operation that only really makes sense on the dosage matrix. The general (multiallelic) construction (which one-hot encode alleles) doesn't really have an analogue.

I don't *think* this impacts the results much---it changes the eigenvalues by a scalar, eigenvectors are essentially the same, on the lostruct "example" we ship---but it does make the two forms numerically distinct. EDIT: that's not the full picture---it does look very similar on the genealogical test bed, but for the highly artificial test fixture with "frequency block structure" they look pretty different. Not sure what the best choice is here.

nspope:
ok this is ready for another look @andrewkern --- I updated the PR description in the OP

andrewkern (approved):
this looks good to go Nate-- we'll want to add a bit in the docs explaining the choice to diverge from the lostruct  R implementation, but we could do that on a separate  PR if you like

nspope:
I think we'll also need to benchmark the two implementations on some real data, unfortunately---these changes have had a bigger blast radius than anticipated.

PR #145, opened 2026-07-14
Multiallelic support for iHS and nSL selection statistics

### LINKED ISSUE #140: Multiallelic support for selection statistics (iHS, nSL)
iHS and nSL assume a biallelic focal SNP. Their homozygosity core is allele-agnostic (it decays on the equality a1 == a2, independent of allele labels), but the focal split partitions haplotypes into ancestral (h == 0) and derived (h == 1) to form the two integrated-homozygosity values whose ratio is the score. On a multiallelic focal site a haplotype carrying an allele index >= 2 falls into neither class and is dropped, and the shared allele-count / MAF filter (_count_alleles_kernel) counts only values 0 and 1, so a common allele with index >= 2 is invisible to min_maf. The windowed mean_nsl statistic calls selection.nsl and inherits this. Garud H, EHH, xpEHH, and xpnSL are allele-agnostic and unaffected. GenotypeMatrix input is biallelic by construction.

#### Generalization

Because the homozygosity core is already allele-agnostic, the fix only changes the focal grouping: one-vs-rest per non-ancestral allele. For each derived allele a at the focal site, compute the integrated homozygosity for haplotypes carrying a versus the reference group, giving a score per (site, allele); a site with one derived allele reduces exactly to the current value. There is no oracle (scikit-allel ihs and nsl are biallelic-only and tskit has no iHS/nSL), so this is a defined extension validated by biallelic reduction and a brute-force internal reference, not tool parity. Two consequences: a multiallelic focal site yields multiple scores (one per derived allele) where today there is one per site, and the multiallelic value is a defensible definition rather than a validated-against-tool number.

#### Fix that is needed regardless

The allele-count / MAF filter must count per allele so min_maf uses the correct per-allele frequencies; otherwise a common allele with index >= 2 is invisible and the wrong sites are analyzed. This is independent of the focal-grouping decision.

#### Decisions deferred to implementation

The reference group for the one-vs-rest split (allele a versus the ancestral allele only, or allele a versus not-a; both reduce to biallelic); the output shape for multiallelic focal sites (one row per (site, allele), or one score per site by aggregating over derived alleles, chosen so the biallelic output is unchanged); and confirming that standardize_by_allele_count bins each per-allele score by that allele's derived count. If the output-shape change is judged too disruptive, the fallback is to biallelic-restrict the focal site with the shared BiallelicOnlyWarning, but the MAF-filter fix is still needed.

andrewkern:
## Summary

Closes #140. Adds multiallelic support to the `iHS` and `nSL` selection statistics. Targets `nsp-multiallelic-trunk`.

iHS and nSL assumed a biallelic focal SNP: the ancestral/derived split **dropped** (iHS) or **merged** (nSL) haplotypes carrying an allele index `>= 2`, and the shared allele-count/MAF filter (`_count_alleles_kernel`) counted only alleles 0 and 1, so a common high-index allele was invisible to `min_maf`.

The shared-haplotype homozygosity core is already allele-agnostic (it decays on `h1 != h2`), so only the **focal grouping** changes: each derived allele is scored **one-vs-ancestral**.

## What changed

- **`pg_gpu/selection.py`**
  - Both focal CUDA kernels gain a `focal_allele` parameter: class 0 = ancestral (`h == 0`), class 1 = the focal derived allele (`h == focal_allele`); a homozygous pair of any other allele extends the shared run but scores neither. `focal_allele = 1` reproduces the old kernels bit-for-bit.
  - `nsl`/`ihs` loop over derived alleles and return the same `(n_variants,)` array as before on biallelic data, or `(n_variants, K-1)` when the matrix has `K > 2` alleles (column `j` = score for derived allele `j+1`; NaN where that allele or the ancestral allele is absent, or the allele is MAF-filtered).
  - MAF now counts **per allele** via `_memutil.allele_counts` (the bespoke 0/1-only `_count_alleles_kernel` is deleted): `min_maf` gates on each derived allele's own frequency `min(freq_a, 1 - freq_a)`, reducing exactly to `min(n0,n1)/(n0+n1)` on biallelic sites.
- **`pg_gpu/windowed_analysis.py`** — `mean_nsl` averages every finite `(site, allele)` score in a window (accepts the 1-D or 2-D `nsl` output).
- `xpEHH`, `XP-nSL`, `EHH decay`, and `Garud H` are allele-agnostic and untouched.

Everything reduces **exactly** to the classic biallelic result.

## Validation

No external oracle exists (scikit-allel `ihs`/`nsl` are biallelic-only; tskit has no iHS/nSL), so `tests/test_selection_multiallelic.py` validates via:

1. **Biallelic reduction** — 1-D output matching the scikit-allel oracle (`rtol=1e-5`), including on `apply_biallelic_filter()` of the multiallelic fixture.
2. **Relabel invariance** — recoding the single derived allele `1 -> 2` moves the score to column 1, leaves column 0 all-NaN, and matches the biallelic score to `atol=1e-9`.
3. **Brute-force reference** — a pure-numpy one-vs-ancestral nSL scan reproduces the GPU per-allele scores on genuine triallelic data.
4. **Per-allele MAF** — a common allele with index `>= 2` is now visible to `min_maf`.
5. **Reference-absent** sites and first-ever **`mean_nsl`** coverage.

All existing selection/windowed tests stay green; full-repo ruff is clean.

## Follow-up (not in this PR)

The per-allele wrapper reruns the full SSL scan once per derived allele — an `M`x factor on multiallelic data (gated on the global max allele index). A single-pass kernel depositing into all allele classes at once would remove it. The biallelic path is one iteration and unaffected.

### DIFF
--- a/pg_gpu/selection.py
+++ b/pg_gpu/selection.py
@@ -321,6 +321,14 @@ def nsl(haplotype_matrix: HaplotypeMatrix,
         is absent at a site). ``standardize_by_allele_count`` takes 1-D input,
         so standardize each derived-allele column separately, pairing column
         ``j`` with that allele's per-site derived count.
+
+    Notes
+    -----
+    nSL scans the shared-haplotype length in both directions across the whole
+    input, so scores depend on the full region. Subsetting or chunking the
+    variants (e.g. windowing via ``mean_nsl``, or a StreamingHaplotypeMatrix)
+    truncates the scan and changes the scores; materialize the region eagerly
+    for a chromosome-wide result.
     """
     if population is not None:
         matrix = _get_population_matrix(haplotype_matrix, population)
@@ -476,6 +484,13 @@ def ihs(haplotype_matrix: HaplotypeMatrix,
         is absent, or the allele is MAF-filtered). ``standardize_by_allele_count``
         takes 1-D input, so standardize each derived-allele column separately,
         pairing column ``j`` with that allele's per-site derived count.
+
+    Notes
+    -----
+    iHS integrates EHH outward from each focal site across the whole input, so
+    scores depend on the full region. Subsetting or chunking the variants (e.g.
+    windowing, or a StreamingHaplotypeMatrix) truncates the scan and changes the
+    scores; materialize the region eagerly for a chromosome-wide result.
     """
     if population is not None:
         matrix = _get_population_matrix(haplotype_matrix, population)
--- a/pg_gpu/selection.py
+++ b/pg_gpu/selection.py
@@ -12,6 +12,7 @@ import cupy as cp
 from typing import Union, Optional, Tuple
 from .haplotype_matrix import HaplotypeMatrix
 from ._utils import get_population_matrix as _get_population_matrix
+from ._memutil import allele_counts
 
 
 # ---------------------------------------------------------------------------
@@ -295,8 +296,10 @@ def nsl(haplotype_matrix: HaplotypeMatrix,
         missing_data: str = 'include'):
     """Compute the unstandardized nSL statistic for each variant.
 
-    Compares the mean shared haplotype length around the reference (0)
-    vs alternate (1) allele at each site.
+    Compares the mean shared haplotype length around the ancestral (0) vs each
+    derived allele at a site, one-vs-ancestral. On a biallelic site this is the
+    classic nSL. On a multiallelic focal site each derived allele gets its own
+    score.
 
     Parameters
     ----------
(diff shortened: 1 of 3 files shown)


nspope (approved):
Looks great! Does mean_nsl need to be added to test_implementation_parity.py ?

andrewkern:
so the situation with nsl is a bit different. There is currently only one implementation, so we don't need a test in `test_implementation_parity.py` _but_ there are two paths that dispatch the kernel, one that is streaming/blocked and one that takes the whole genotype matrix. the streaming path truncates the snps summed over by design so the two paths aren't expected to be equivalent. We might add a loud warning to the streaming mean_nsl path?

nspope:
Got it. I'm not sure a warning is the right choice because it'd always trigger, rather than flagging unusual behaviour (right?) So maybe just making the difference explicitly obvious in the docstring? If it's not already?

Or wait, it'll trigger depending on dispatch type, which is automatic but can be overriden? In this case, yes a warning seems like the right choice. But happy to defer here---whatever you think is best.

andrewkern:
what do you think about the "one-vs-ancestral" choice for the multi-allelic case? I think this is the right thing to do?

nspope:
It's the right choice, better than the "particular allele versus all others" I had originally proposed. It's technically "particular allele versus reference" though, right? Should this be made clear (that data need to be polarised)?

andrewkern:
yes indeed

nspope:
i'm going to go ahead and merge, we'll revisit docs in an editing pass later

PR #149, opened 2026-07-16
Multiallelic support for grm and ibs; new genetic_relatedness that is general haplotype form

nspope:
**Background** relatedness reconstructed diploids from haplotypes by summing paired allele indices, so on a site with an allele index >= 2 the "dosage" became 3, 4, ... and the GRM's frequency and standardization went out of range. This PR makes relatedness multiallelic-correct by reframing it around what tskit actually computes. `genetic_relatedness` is a new statistic pinned exactly to `tskit.genetic_relatedness`; `grm` and `ibs` become the two diploid biallelic estimators scoped to `GenotypeMatrix`.

#### What changed

Added `genetic_relatedness(haplotype_matrix, sample_sets, indexes, *, centre, polarised, proportion, span_normalize, missing_data)`: the site-mode centered per-allele allele-frequency covariance between sample sets. For sets i and j it sums over sites and alleles `(p_i(a) - pbar(a)) * (p_j(a) - pbar(a))`, where `p_k(a)` is allele a's frequency in set k and `pbar(a)` the unweighted mean across sets, with no variance standardization. As a matrix this is `C @ C.T` where `C` holds the per-set per-allele frequencies centered by `pbar`. It is multiallelic-correct by construction (it is already per-allele) and one computation covers every grouping: one set per haplotype (the default) gives the sample-by-sample matrix, two-haplotype sets give per-individual, and population sets give per-population relatedness. It pins exactly to tskit for centre, polarised, proportion, and span_normalise; proportion=True divides by the segregating-site count and, as in tskit, then makes span_normalize a no-op.

`grm` is now scoped to `GenotypeMatrix` (and `StreamingGenotypeMatrix`), which are diploid and biallelic by construction, and keeps the GCTA convention there unchanged. A `HaplotypeMatrix` is hard-rejected with a message pointing to `genetic_relatedness` (or to `GenotypeMatrix.from_haplotype_matrix` for the GCTA GRM). This surfaced and fixed a pre-existing bug where `grm(GenotypeMatrix, population=...)` raised, because `get_population_matrix` only indexed haplotypes; it now subsets a `GenotypeMatrix` too.

`ibs` is also scoped to `GenotypeMatrix` and stays the diploid biallelic PLINK IBS (per-site shared count `2 - abs(g_i - g_j)`, summed over jointly-called sites over twice the count); it hard-rejects haplotype input like `grm`. It is deliberately not generalized to haplotypes or multiallelic sites: the only coherent sample-set allele-sharing quantity, sum over alleles of `p_i(a) p_j(a)`, is exactly the noncentred `genetic_relatedness`, so a general `ibs` would only re-expose that. The docstring points users wanting the general or multiallelic allele sharing to `genetic_relatedness` with `centre=False`.

Streaming: `genetic_relatedness` streams the variant axis in one pass and accumulates `C @ C.T` into a host matrix tiled on the sample axis, the same host-accumulate plus row-block scheme as the streaming GRM/IBS, so it holds when the output is large (the per-haplotype default). It reuses `iter_gpu_chunks` and the existing block-size estimators.

#### Verification

`genetic_relatedness` is pinned exactly to `ts.genetic_relatedness` on a multiallelic tree sequence across singleton and grouped sample sets, centre True and False, polarised True and False, proportion, span-normalized and raw, and index selection; missing include/exclude is checked against an independent host reference (pg_gpu's own extension, since tree sequences have no missing data). `grm` keeps a biallelic GCTA host reference; `ibs` keeps a biallelic dosage host reference (including missing data); both add a haplotype-rejection test. Streaming equals eager for all three, with chunk-boundary invariance.

### DIFF
--- a/pg_gpu/_utils.py
+++ b/pg_gpu/_utils.py
@@ -6,38 +6,49 @@ from typing import Union
 from .haplotype_matrix import HaplotypeMatrix
 
 
-def get_population_matrix(haplotype_matrix: HaplotypeMatrix,
-                          population: Union[str, list]) -> HaplotypeMatrix:
-    """Extract a population-specific HaplotypeMatrix.
+def get_population_matrix(matrix, population: Union[str, list]):
+    """Extract a population-specific subset matrix.
 
     Parameters
     ----------
-    haplotype_matrix : HaplotypeMatrix
-        The full haplotype data.
+    matrix : HaplotypeMatrix or GenotypeMatrix
+        The full data. Rows are haplotypes (HaplotypeMatrix) or individuals
+        (GenotypeMatrix); the subset is taken along that row axis.
     population : str or list
-        Population name (looked up in sample_sets) or list of sample indices.
+        Population name (looked up in sample_sets) or list of row indices.
 
     Returns
     -------
-    HaplotypeMatrix
-        Subset matrix for the specified population.
+    HaplotypeMatrix or GenotypeMatrix
+        Subset matrix of the same type for the specified population.
     """
+    from .genotype_matrix import GenotypeMatrix
+
     if isinstance(population, str):
-        if haplotype_matrix.sample_sets is None:
-            raise ValueError("No sample_sets defined in haplotype matrix")
-        if population not in haplotype_matrix.sample_sets:
+        if matrix.sample_sets is None:
+            raise ValueError("No sample_sets defined in matrix")
+        if population not in matrix.sample_sets:
             raise ValueError(
                 f"Population {population} not found in sample_sets")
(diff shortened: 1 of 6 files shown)


andrewkern:
Reviewed locally — LGTM. One low-severity follow-up, tracked in #151 (with an MRE): streaming `genetic_relatedness` with the default `span_normalize=True` normalizes by the raw per-base span and does not honor an accessible mask / `n_total_sites` the way the eager path does, so eager and streaming can diverge on that path when a mask is set. The raw (`span_normalize=False`) and `proportion=True` paths are unaffected.

Issue #152, opened 2026-07-21
GenotypeMatrix streaming loader drops accessible_bed: streaming grm/ibs ignore the accessible mask


andrewkern:
## Summary

`GenotypeMatrix.from_zarr(..., streaming=...)` accepts `accessible_bed` but the streaming branches never forward it, so a `StreamingGenotypeMatrix` carries no accessible mask. Eager `grm`/`ibs` restrict to accessible variants (an accessible mask filters the `GenotypeMatrix.genotypes` view), but streaming `grm`/`ibs` use *all* variants — a silent eager-vs-streaming divergence under an `accessible_bed`.

This is the same drop that #151 fixed for `HaplotypeMatrix` (and streaming `genetic_relatedness`), still present for the sibling class.

## Root cause

- `GenotypeMatrix.from_zarr` (`genotype_matrix.py:432`) takes `accessible_bed`, but both `_build_streaming` calls forward everything *except* it (`genotype_matrix.py:482-486`, `:504-508`); only the eager branch forwards it (`:509-511`).
- `StreamingGenotypeMatrix` inherits `accessible_mask` / `get_span` from `_StreamingMatrixBase` (added in #151) but nothing ever populates them, so it always falls back to raw / no-mask.
- `grm`/`ibs` require a `GenotypeMatrix`, so they only ever see the unmasked streaming matrix. Eager `grm`/`ibs` filter to accessible sites via the mask-filtered `.genotypes` view (`relatedness.py` `_get_genotype_data`).

Note: `grm`/`ibs` are per-site averages, not span-normalized, so — unlike `genetic_relatedness` — this is purely about **variant filtering**, not the span denominator.

## Impact

`grm`/`ibs` on a `from_zarr(streaming="always", accessible_bed=bed)` matrix silently include inaccessible sites, diverging from the eager result on the same data. No error is raised.

## Proposed fix

Mirror #151 for `GenotypeMatrix`:

1. Forward `accessible_bed` from both `GenotypeMatrix.from_zarr` streaming branches (`genotype_matrix.py:482`, `:504`) into `GenotypeMatrix._build_streaming`.
2. In `_build_streaming`, resolve the mask once via `resolve_accessible_mask(accessible_bed, source.mappable_lo, source.mappable_hi - 1, chrom)` and pass it to `StreamingGenotypeMatrix(..., accessible_mask=...)` — the same block `HaplotypeMatrix._build_streaming` now uses.

[...]

PR #153, opened 2026-07-21
Streaming genetic_relatedness: honor the accessible mask (span + variant filtering)

### LINKED ISSUE #151: Streaming span-normalization ignores accessible-span metadata (mask / n_total_sites), diverging from eager
## Summary

Span-normalized statistics computed on a **streaming** matrix silently normalize by the raw per-base span (`chrom_end - chrom_start + 1`), ignoring the accessible-span metadata (an accessible mask, or `n_total_sites`) that the **eager** path honors via `HaplotypeMatrix.get_span('auto')`. The two paths therefore return different values for the *same data* on the default `span_normalize=True` path whenever the eager matrix carries a mask / `n_total_sites`.

The concrete instance is `genetic_relatedness` (introduced in the `nsp-multiallelic-6` branch / #149), but the root cause is general: streaming matrices carry no accessible-span concept, so any span-normalized streaming reduction inherits the gap.

## Root cause

- Eager `genetic_relatedness` normalizes by `haplotype_matrix.get_span(mode)` — priority **accessible mask > `n_total_sites` > raw per-base span** (`pg_gpu/relatedness.py:122`, `HaplotypeMatrix.get_span`).
- Streaming `_stream_genetic_relatedness` hardcodes `streaming_matrix.chrom_end - streaming_matrix.chrom_start + 1` (`pg_gpu/relatedness.py:194`).
- `_StreamingMatrixBase` exposes no `get_span`, `accessible_mask`, or `n_total_sites` (grep count 0 in `pg_gpu/streaming_matrix.py`), and the streaming loaders take no `accessible_bed`, so streaming has no way to express an accessible span.

The raw (`span_normalize=False`) matrices are identical between the two paths — the algorithm and the streaming accumulation are correct. The divergence is purely the denominator.

## MRE

```python
import os, tempfile, numpy as np
from pg_gpu import HaplotypeMatrix, relatedness

hap = np.random.RandomState(0).randint(0, 2, (8, 60)).astype(np.int8)
pos = np.sort(np.random.RandomState(1).choice(range(1, 100_000), 60, replace=False)).astype(np.int64)
path = os.path.join(tempfile.mkdtemp(), "mre.vcz")
HaplotypeMatrix(hap, pos, 0, 99_999).to_zarr(path, format="vcz", contig_name="1")

eager  = HaplotypeMatrix.from_zarr(path, streaming="never")
stream = HaplotypeMatrix.from_zarr(path, streaming="always", chunk_bp=10_000)
eager.n_total_sites = 40_000     # accessible bases < raw span (what an accessible_bed load sets)

raw_agree = np.allclose(relatedness.genetic_relatedness(eager,  span_normalize=False),
                        relatedness.genetic_relatedness(stream, span_normalize=False))
R_e = relatedness.genetic_relatedness(eager,  span_normalize=True)   # divides by 40000
R_s = relatedness.genetic_relatedness(stream, span_normalize=True)   # divides by raw span
print("raw matrices agree:", raw_agree)                    # True
print("normalized max abs diff:", np.abs(R_e - R_s).max()) # ~4.6e-4
print("ratio R_e/R_s[0,1]:", R_e[0, 1] / R_s[0, 1])        # ~2.48 == raw_span / 40000
```

Output:
```
raw matrices agree: True
normalized max abs diff: 0.0004564...
ratio R_e/R_s[0,1]: 2.4835
```

## Impact / scope

- Silent: no error, just a different-by-a-constant relatedness matrix depending on load mode. `proportion=True` and `span_normalize=False` are unaffected (they don't use the span, or it cancels).
- Also affects the docstring claim that eager `genetic_relatedness` "matches tskit's `span_normalise`": tskit divides by sequence length, whereas the eager pg_gpu path divides by *accessible bases* when a mask is present — so the eager path already differs from tskit under a mask, independently of streaming.

## Proposed fix

Pick one:

1. **Plumb accessible-span metadata through streaming (most correct).** Let the streaming loaders accept `accessible_bed` / `n_total_sites` and give `_StreamingMatrixBase` a `get_span('auto')` mirroring the eager one; route `_stream_genetic_relatedness` (and any future span-normalized streaming stat) through it. Larger change, but closes the gap generally.

2. **Consult the metadata if present (small).** In `_stream_genetic_relatedness`, before the raw-span fallback, check `getattr(m, 'accessible_mask', None)` then `getattr(m, 'n_total_sites', None)` (a shared helper mirroring `get_span('auto')`). Makes eager == streaming whenever the info is set; plain streaming matrices are unchanged (raw span, which already matches a mask-free eager matrix).

3. **Document + guard (smallest).** Document that streaming span-normalization uses the raw per-base span only and cannot honor an accessible mask; leave normalization to the caller or recommend materializing eagerly for mask-aware relatedness.

In all cases the eager docstring's "matches tskit's `span_normalise`" wording should be corrected to note the accessible-bases convention.

andrewkern:
## Summary

Closes #151. Makes streaming `genetic_relatedness` honor an accessible mask the same way the eager path does.

Eager `genetic_relatedness` normalizes by `get_span('auto')` (accessible-mask count > `n_total_sites` > raw per-base span) and, because an accessible mask filters a `HaplotypeMatrix`'s variant view, computes over accessible sites only. The streaming path did neither: `accessible_bed` was dropped by `from_zarr`'s streaming branches, `StreamingHaplotypeMatrix` carried no mask, and `_stream_genetic_relatedness` hardcoded the raw chunk-grid span. So the same data loaded eagerly vs streamed diverged on the default `span_normalize=True` path whenever a mask was present (raw matrices identical, normalized differ by `raw_span / accessible_bases`).

## What changed

- **`streaming_matrix.py`** — `_StreamingMatrixBase` gains `accessible_mask` / `n_total_sites` and a `get_span('auto')` mirroring `HaplotypeMatrix.get_span`, using the source's variant-position bounds (`mappable_lo`/`mappable_hi`) so the span matches an eager matrix built from the same store. `iter_gpu_chunks` attaches the mask to every chunk it yields — the single chunk-production point — so **every** streaming consumer (`genetic_relatedness`, and `windowed_analysis`) sees accessible-filtered variants by construction, rather than each reduction re-applying the mask.
- **`haplotype_matrix.py`** — `from_zarr` forwards `accessible_bed` to `_build_streaming`, which resolves the mask once and attaches it.
- **`relatedness.py`** — `_stream_genetic_relatedness` takes its span from `get_span` and drops its own per-chunk attach (now redundant given the choke-point); docstrings corrected.

## Verification

- New tests: eager/streaming `genetic_relatedness` parity with and without an `accessible_bed` (both filtering and span), and that `iter_gpu_chunks` filters chunk variants to accessible sites.
- The #151 MRE is now exact: `max|eager − stream| == 0` with and without a mask, identical variant sets.
- Full relatedness + streaming + zarr + accessible-mask suites stay green; full-repo ruff clean.

## Follow-up (not in this PR)

The sibling `StreamingGenotypeMatrix` loader still drops `accessible_bed`, so streaming `grm`/`ibs` ignore a mask (variant filtering, not span, since they're per-site averages). Tracked in #152; the `iter_gpu_chunks` choke-point here means the fix there is just plumbing the mask onto the genotype streaming matrix.

### DIFF
--- a/pg_gpu/haplotype_matrix.py
+++ b/pg_gpu/haplotype_matrix.py
@@ -626,7 +626,7 @@ class HaplotypeMatrix:
             return cls._build_streaming(
                 path, region=region, pop_assignment=pop_assignment,
                 chunk_bp=chunk_bp, prefetch=prefetch,
-                backend=backend,
+                backend=backend, accessible_bed=accessible_bed,
             )
 
         # 'auto' and 'never' both want eager when the matrix fits; 'auto'
@@ -649,7 +649,7 @@ class HaplotypeMatrix:
             return cls._build_streaming(
                 path, region=region, pop_assignment=pop_assignment,
                 chunk_bp=chunk_bp, prefetch=prefetch,
-                source=source, backend=backend,
+                source=source, backend=backend, accessible_bed=accessible_bed,
             )
         return cls._build_eager(path, region=region,
                                 accessible_bed=accessible_bed,
@@ -716,7 +716,8 @@ class HaplotypeMatrix:
 
     @classmethod
     def _build_streaming(cls, path, *, region, pop_assignment, chunk_bp,
-                         prefetch, source=None, backend="auto"):
+                         prefetch, source=None, backend="auto",
+                         accessible_bed=None):
         from .streaming_matrix import (
             HostChunkFetcher, KvikioChunkFetcher, StreamingHaplotypeMatrix,
             _pick_chunk_fetcher,
@@ -733,9 +734,24 @@ class HaplotypeMatrix:
             # zarr store.
             source.pop_cols = source._resolve_pop_assignment(pop_assignment)
         fetcher = _pick_chunk_fetcher(source, backend=backend)
+
+        # Resolve an accessible BED once over the source's variant-position
+        # bounds and hand it to the streaming matrix, so span-normalized
+        # reductions (genetic_relatedness) divide by accessible bases the same
+        # way the eager path does. ``mappable_hi`` is one past the last variant,
+        # so the inclusive last position is ``mappable_hi - 1``.
+        accessible_mask = None
+        if accessible_bed is not None:
+            from .accessible import resolve_accessible_mask
+            chrom = region.split(':')[0] if region else source.chrom
+            lo, hi = source.mappable_lo, source.mappable_hi
+            accessible_mask = resolve_accessible_mask(
+                accessible_bed, lo, hi - 1, chrom)
+
         return StreamingHaplotypeMatrix(
             source, fetcher,
             chunk_bp=chunk_bp, prefetch=prefetch,
+            accessible_mask=accessible_mask,
         )
 
     def to_zarr(self, zarr_path: str, format: str = 'vcz',
--- a/pg_gpu/relatedness.py
+++ b/pg_gpu/relatedness.py
@@ -48,7 +48,11 @@ def genetic_relatedness(haplotype_matrix, sample_sets=None, indexes=None, *,
         between the covariance and the segregating-site count).
     span_normalize : bool or str
         True (default) divides by the analysis span (see ``get_span``); False
-        returns the raw sum. Matches tskit's ``span_normalise``.
+        returns the raw sum. Like tskit's ``span_normalise``, except that when
+        the matrix carries an accessible mask the span is the accessible-base
+        count over the analyzed region rather than the raw sequence length. The
+        eager and streaming paths use the same span source (an ``accessible_bed``
+        supplied at load).
     missing_data : str
         'include' (default) uses per-site, per-set valid counts; 'exclude'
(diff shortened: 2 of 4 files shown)


andrewkern:
Adversarial review pass found one bug (now fixed in the pushed update): the streaming `get_span` returned the *unfiltered* variant count/span for the `per_variant`/`sites`/`callable` modes, so under an accessible mask those non-default `span_normalize` modes disagreed with eager (the numerator is accumulated over accessible-filtered chunks). It now computes those over the accessible-filtered variant set from `source.site_pos`, matching eager; `per_base`/`total` stays the raw span. Added a parametrized test over the modes. The default `span_normalize=True` ('auto') path was already correct. Everything else reviewed clean (coordinate reconciliation, the iter_gpu_chunks choke-point vs the windowed call-time mask override, whole-region mask on sub-region chunks).

PR #155, opened 2026-07-21
Multiallelic support for decomposition.pairwise_distance

nspope:
#### Background

decomposition.pairwise_distance built its euclidean / sqeuclidean / cityblock distances from the raw allele-index matrix, using the allele index as a magnitude. On a multiallelic site that makes the distance depend on arbitrary allele numbering (|allele 3 - allele 0| = 3 but |allele 2 - allele 1| = 1), the same index-inflation class of bug as the old PCA path. It feeds pcoa (classical MDS). GenotypeMatrix input previous errored out with an obscure message (no attribute "haplotypes"), and now raises a TypeError.

#### What changed

The three supported metrics are now built from the allele-mismatch count m: the number of jointly non-missing sites at which two haplotypes carry different alleles. cityblock and sqeuclidean are m; euclidean is sqrt(m). m is label-independent, so the distance no longer depends on how alleles are numbered, and it is correct at multiallelic sites. On categorical (allele-index) data the three metrics are all functions of the single quantity m, so they coincide up to that transform.

pairwise_distance now routes its mismatch count through the shared one-hot Hamming kernel distance_stats._pairwise_diffs_matrix_gpu (the multiallelic-correct kernel already used by pairwise_diffs_haploid and the divergence two-population distances), rather than carrying its own per-metric computation. The kernel gains an additive return_joint_valid flag (default off, so existing callers are byte-for-byte unchanged) so pairwise_distance can apply the exact prior normalization m * n_var / n_joint (multiply-before-divide) over the jointly non-missing site count from a single pass.

The scipy fallback for other metrics is removed. It was unused (no caller or test exercised it) and ran on the raw allele-index matrix. Most scipy metrics are magnitude-based (euclidean-like, chebyshev, correlation, ...) and index-inflate on multiallelic input; a few (hamming, matching) are label-independent and would be correct as-is, and the boolean metrics collapse multiallelic to presence/absence. Rather than generalize twenty metrics case-by-case for a dead path, an unsupported metric now raises NotImplementedError.

Passing GenotypeMatrix would previously have results in an ambiguous error, now it raises a TypeError that explicitly points to HaplotypeMatrix.

#### Behaviour

Biallelic output is unchanged, bit-for-bit, including the missing-data normalization path. On multiallelic data the distances are now label-independent instead of index-inflated. pcoa is unaffected: when it is called on pairwise_distance output the input distances change on multiallelic data (correctly), and its other consumer, lostruct, feeds it pc_dist output rather than pairwise_distance, so that path is untouched. The memory profile of pairwise_distance changes from pair-batching to the kernel's variant-chunked (n_hap, n_hap) accumulation, matching pairwise_diffs_haploid at the same sample count.

#### Verification

- Biallelic bit-identity, bit-for-bit (array_equal) against a reconstruction of the pre-multiallelic per-pair formula, across clean and both missing-data modes for all three metrics -- this covers the missing-data normalization path, which scipy cannot oracle since it treats missing (-1) as a value. The pre-existing biallelic-vs-scipy tests (clean, rtol) still pass.
- Multiallelic parity against an independent host mismatch-count reference, clean and missing, all three metrics.
- Label-independence: permuting allele labels leaves the distances unchanged.
- Reduction: a 2-allele site coded {0,2} gives the same distances as {0,1}.
- Metric collapse: cityblock == sqeuclidean, euclidean == sqrt, on categorical data.
- Unsupported metric raises.
- The shared kernel's existing callers (pairwise_diffs_haploid, the divergence two-population distances, the diploSHIC distance features) are unchanged.
- Passing a GenotypeMatrix raises a type error.

### DIFF
--- a/pg_gpu/decomposition.py
+++ b/pg_gpu/decomposition.py
@@ -394,18 +394,27 @@ def pairwise_distance(haplotype_matrix: HaplotypeMatrix,
                       missing_data: str = 'include'):
     """Compute pairwise distances between samples.
 
+    Distances are built from the allele-mismatch count m: the number of sites at
+    which two haplotypes carry different alleles (over jointly non-missing sites).
+    This is label-independent, so it does not depend on how alleles are numbered
+    and is correct at multiallelic sites. cityblock and sqeuclidean are both m;
+    euclidean is sqrt(m). On categorical data these coincide up to that transform,
+    and on biallelic data m equals the usual per-site difference count, so the
+    values match the standard distances there.
+
     Parameters
     ----------
     haplotype_matrix : HaplotypeMatrix
         Haplotype data.
     metric : str
-        Distance metric. Supported on GPU: 'euclidean', 'cityblock',
-        'sqeuclidean'. Falls back to scipy for other metrics.
+        Distance metric: 'euclidean' (default), 'cityblock', or 'sqeuclidean'.
+        Other metrics raise NotImplementedError.
     population : str or list, optional
         Population subset.
     missing_data : str
-        'include' - mask missing, normalize per pair
-        'exclude' - filter sites with any missing
+        'include' - count mismatches over jointly non-missing sites, scaled to
+        the full variant span per pair
+        'exclude' - filter sites with any missing first
 
     Returns
     -------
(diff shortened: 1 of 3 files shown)


andrewkern (approved):
this looks great @nspope -- this might even lead to a performance speed up?

PR #158, opened 2026-07-21
Streaming GenotypeMatrix honors accessible_bed for grm/ibs (#152)

### LINKED ISSUE #152: GenotypeMatrix streaming loader drops accessible_bed: streaming grm/ibs ignore the accessible mask

andrewkern:
Closes #152.

## Problem

`GenotypeMatrix.from_zarr(..., streaming=..., accessible_bed=bed)` accepted `accessible_bed` but both streaming branches dropped it, so a `StreamingGenotypeMatrix` carried no accessible mask. `grm`/`ibs` then streamed over **all** variants while the eager path filtered to accessible sites via the mask-filtered `.genotypes` view -- a silent eager-vs-streaming divergence under an `accessible_bed`, with no error raised. This is the sibling of #151, which fixed the same drop for `HaplotypeMatrix`.

## Fix

- Forward `accessible_bed` from both `GenotypeMatrix.from_zarr` streaming branches into `_build_streaming`, resolve the mask once over the source's variant bounds, and hand it to `StreamingGenotypeMatrix`. `iter_gpu_chunks` already attaches the mask to every chunk, so per-chunk variant filtering happens for free and streaming `grm`/`ibs` match eager.
- Factor the shared "resolve mask over source bounds" block into `resolve_streaming_accessible_mask`, now used by both the haplotype and genotype streaming loaders.
- Correct the stale `StreamingGenotypeMatrix` docs/error text, which still claimed `grm` needs `.materialize()` -- it streams directly.

## Audit

Swept the whole streaming surface for this bug class (mask honored on eager, dropped on streaming):

- **Loaders**: only `from_zarr` streams; both `HaplotypeMatrix` (#151) and `GenotypeMatrix` (this PR) now forward the mask. `from_vcf`/`from_ts` are eager-only.
- **Consumers**: no stats module reads raw arrays -- all go through the accessible-filtered `.haplotypes`/`.genotypes`/`.positions` properties. Verified per streaming consumer: `sfs`, `genetic_relatedness` (#153), `grm`/`ibs`, `windowed_analysis`, and `ld_statistics` all honor the mask. `selection`, `admixture`, `diversity`, `divergence`, `distance_stats`, `ld_statistics_genotype`, and `local_pca` have no streaming-matrix input path (eager-only or raise).

Conclusion: no other statistic carried a live instance of this bug. The one residual was a test-coverage gap for the load-time-mask -> streaming `windowed_analysis` path, closed here with a regression test.

## Tests

- Eager-vs-streaming `grm`/`ibs` parity with and without an `accessible_bed`, plus a guard that the mask actually changes the streaming result and a chunk-level check that each streaming chunk is restricted to accessible positions.
- Pin the load-time mask through streaming `windowed_analysis`.
- Full suite green (`test_streaming_kernels`, `test_accessible_mask`, `test_relatedness`, matrix/streaming loaders); `ruff` clean.

### DIFF
--- a/tests/test_streaming_kernels.py
+++ b/tests/test_streaming_kernels.py
@@ -613,3 +613,34 @@ class TestGrmIbsAccessibleBed:
             assert ((pos >= 25000) & (pos < 75000)).all()
             seen += chunk.num_variants
         assert seen > 0
+
+    @pytest.mark.parametrize("stat", ["grm", "ibs"])
+    def test_auto_fallback_forwards_accessible_bed(self, vcz_store, tmp_path,
+                                                   monkeypatch, stat):
+        # streaming='auto' forwards accessible_bed on its streaming-fallback
+        # branch too, not just streaming='always'. The tiny fixture fits
+        # eagerly so 'auto' never falls back on its own; force the fallback
+        # by making the size decision return 'streaming' with a real source,
+        # which drives the actual _build_streaming(source=..., accessible_bed=)
+        # call. Guards against dropping accessible_bed from only that branch.
+        import pg_gpu.haplotype_matrix as hm_mod
+        from pg_gpu.zarr_source import ZarrGenotypeSource
+        from pg_gpu.streaming_matrix import StreamingGenotypeMatrix
+        from pg_gpu import relatedness
+        fn = getattr(relatedness, stat)
+        bed = self._write_bed(str(tmp_path / "acc.bed"))
+
+        def force_streaming(zarr_path, region, streaming, pop_assignment, **kw):
+            src = ZarrGenotypeSource(zarr_path, region=region,
+                                     pop_assignment=False)
+            return "streaming", src
+        monkeypatch.setattr(hm_mod, "_decide_streaming_mode", force_streaming)
+
+        stream = GenotypeMatrix.from_zarr(vcz_store, streaming="auto",
+                                          chunk_bp=10_000, accessible_bed=bed)
+        assert isinstance(stream, StreamingGenotypeMatrix)  # fallback taken
+        assert stream.accessible_mask is not None            # bed forwarded
+        eager = GenotypeMatrix.from_zarr(vcz_store, streaming="never",
+                                         accessible_bed=bed)
+        np.testing.assert_allclose(fn(stream), fn(eager),
+                                   rtol=1e-6, atol=1e-9)
--- a/tests/test_streaming_kernels.py
+++ b/tests/test_streaming_kernels.py
@@ -239,6 +239,35 @@ class TestAccessibleBedDispatch:
             seen += chunk.num_variants
         assert seen > 0
 
+    def test_load_time_mask_windowed_equivalent(self, vcz_store, tmp_path):
+        # A LOAD-TIME mask must reach windowed_analysis without re-passing
+        # accessible_bed: iter_gpu_chunks attaches it, and the per-chunk
+        # eager call honors the already-attached mask. This is a distinct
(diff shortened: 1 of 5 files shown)


tests/test_streaming_kernels.py
```diff
@@ -516,3 +545,102 @@ def test_span_mode_matches_eager_with_mask(self, vcz_store, tmp_path, mode):
         e = relatedness.genetic_relatedness(eager, span_normalize=mode)
         s = relatedness.genetic_relatedness(stream, span_normalize=mode)
         np.testing.assert_allclose(s, e, rtol=1e-9, atol=1e-12)
+
+
+class TestGrmIbsAccessibleBed:
```
nspope, line 599: This is good, but doesn't have a case where there's missing data. It should be easy enough to inject missing data (-1) into the vcz --- I think that's important to cover for the parity check between streaming and eager (cover interaction between missing data (both modes) and the mask).

nspope (approved):
Looks good---one comment about missing test coverage.

nspope:
Went ahead and added this test, merging

PR #160, opened 2026-07-22
Multiallelic-correct windowed daf_hist / mu_sfs

nspope:
The windowed diploSHIC features `daf_hist` and `mu_sfs` derived their derived-allele counts from `dac = sum(max(hap, 0))`, which sums allele indices, so on a multiallelic site the derived-allele frequency was index-inflated and could exceed the sample size. They also diverged from the tskit-pinned scalar `diversity.daf_histogram` / `diversity.mu_sfs` on plain biallelic data: a fixed `n_hap` instead of per-site `n_valid`, raw counts instead of a normalized histogram, a total-variant instead of a segregating denominator, and a missing-data edge misclassification in `mu_sfs`.

This aligns both windowed features, in the fused (`missing_data='include'`) engine, to the scalar functions, so the windowed value over a window equals the scalar computed on that window's variants, across biallelic / multiallelic and clean / missing data. `GenotypeMatrix` input is biallelic by construction and out of scope.

#### What changed

`windowed_statistics_fused` (and its chunked twin, which delegates these stats to it) now computes `daf_hist` and `mu_sfs` per allele via `_memutil.allele_counts` with per-site `n_valid`, matching `diversity`: `daf_hist` bins one frequency per present derived allele (`count / n_valid`) into (window, bin) and normalizes per window; `mu_sfs` counts segregating (`0 < count < n_valid`) and edge (`count == 1` or `count == n_valid - 1`) (site, allele) entries and returns `edge / segregating` (0.0 when a window has none). A shared `diversity._daf_bin_index` makes the windowed scatter and the scalar histogram assign bins identically. The windowed `daf_hist` bin count, previously hardcoded as 20 in two places, is now the single module constant `_DAF_N_BINS`.

This is the include-mode fused fast path. Exclude mode is unsupported for these two stats (requesting them with `missing_data='exclude'` raises `Unknown statistic`, unchanged); adding it is future work.

#### Behaviour

These are now the per-allele definitions of the two features: multiallelic-correct, using `allele_counts` and per-site `n_valid`, the same per-allele convention as the SFS, `max_daf`, and the scalar `diversity.daf_histogram` / `diversity.mu_sfs`. Nothing is restricted to biallelic; a biallelic site is the one-derived-allele special case. Because the features now follow that one consistent per-allele convention rather than the old index-summed / fixed-`n_hap` / raw-count one, their output changes on all `HaplotypeMatrix` input, biallelic included -- an intentional redefinition of the two diploSHIC feature vectors, not a silent bugfix. `daf_hist` becomes a normalized per-window histogram over present derived alleles (no monomorphic zero-frequency spike), and `mu_sfs` uses the segregating denominator with correct per-site edge classification, returning 0.0 rather than NaN for a window with no segregating entries.

#### Verification

`diversity.daf_histogram` / `diversity.mu_sfs` (pinned to tskit's polarised allele frequency spectrum) are the oracle; the contract is that the windowed value equals the scalar over a window. Covered: whole-region windowed equals scalar for biallelic / multiallelic and clean / include-missing, including a site fixed for the derived allele with one missing haplotype (which must not count as an SFS edge); streaming equals eager across chunk and window combinations (include, and include with missing data); and both stats in the cross-implementation parity harness (`mu_sfs` in the scalar matrix, `daf_hist` as a dedicated vector check) for include mode.

#### Bugs noticed in passing

- Window-boundary membership and overlapping windows. The fused scatter assigns each variant to a single window via `searchsorted` on window ends, so a variant sitting exactly on an interior window boundary is scattered into the lower window while `n_variants` / `pi` / theta attribute it to the upper one, and with overlapping windows (`step_size < window_size`) a variant is counted in only one of its windows, so `daf_hist` / `mu_sfs` / `mean_nsl` undercount silently. Pre-existing (the earlier `daf_hist` used the same assignment), and the whole-region parity here does not exercise it. Tracked as a separate windowing issue.
- Streaming with `missing_data='exclude'` diverges from eager at chunk boundaries for all windowed stats (`pi` included), a pre-existing streaming window-grid issue, tracked separately.
- Making the bin count a caller parameter is left as a follow-up.

### DIFF
--- a/pg_gpu/windowed_analysis.py
+++ b/pg_gpu/windowed_analysis.py
@@ -63,8 +63,8 @@ CANONICAL_WINDOW_PREFIX = (
     'chrom', 'start', 'end', 'center', 'n_variants', 'window_id')
 
 # Number of bins in the windowed daf_hist feature (columns daf_bin_0 ..
-# daf_bin_{_DAF_N_BINS - 1}). Fixed for now, shared by the fused engine and the
-# fallback so they emit the same columns; making it a caller parameter is deferred.
+# daf_bin_{_DAF_N_BINS - 1}) emitted by the fused engine. Fixed for now; making
+# it a caller parameter is deferred.
 _DAF_N_BINS = 20
 
 
@@ -233,16 +233,9 @@ class StatisticsComputer:
                                                 span_normalize=sn),
         }
 
-        # Frequency-spectrum features computed here (the exclude-mode path) by
-        # deferring to the scalar diversity functions per window; the fused engine
-        # handles the include-mode fast path. daf_hist is vector-valued and expands
-        # into daf_bin_0 .. daf_bin_{n-1} columns.
-        self.FREQ_STATS = {'daf_hist', 'mu_sfs'}
-
         self.single_pop_stats = []
         self.two_pop_stats = []
         self.ld_stats = []
-        self.freq_stats = []
         self.custom_stats = []
 
         for stat in self.statistics:
@@ -253,8 +246,6 @@ class StatisticsComputer:
                     self.two_pop_stats.append(stat)
                 elif stat in self.LD_STATS:
                     self.ld_stats.append(stat)
-                elif stat in self.FREQ_STATS:
-                    self.freq_stats.append(stat)
                 else:
                     raise ValueError(f"Unknown statistic: {stat}")
             else:
@@ -274,17 +265,12 @@ class StatisticsComputer:
 
         # Skip if no variants
         if window.n_variants == 0:
-            # Empty window: scalar stats are undefined -> NaN. The frequency-
-            # spectrum features are vector-valued (their own daf_bin_* / mu_sfs
-            # columns) and must still emit those columns, filled with the value
-            # the scalar and fused paths give for no data (0.0 / all-zero
-            # histogram) -- handled below by _freq_stat_results.
+            # Fill with NaN for all statistics
             for stat in self.statistics:
-                name = stat if isinstance(stat, str) else stat.__name__
-                if name in self.FREQ_STATS:
-                    continue  # emitted as their own columns below
-                results[name] = np.nan
-            results.update(self._freq_stat_results(window.matrix, 0))
+                if isinstance(stat, str):
+                    results[stat] = np.nan
+                else:
+                    results[stat.__name__] = np.nan
             return results
 
         # Single population statistics
@@ -321,43 +307,12 @@ class StatisticsComputer:
                 kwargs['bins'] = self.ld_bins
             results[stat] = self.LD_STATS[stat](window, **kwargs)
 
-        # Frequency-spectrum features (daf_hist / mu_sfs), computed on the first
-        # population if any, matching the fused engine's population handling.
-        if self.freq_stats:
-            fmatrix = (self._get_population_matrix(window.matrix, self.populations[0])
-                       if self.populations else window.matrix)
-            results.update(self._freq_stat_results(fmatrix, window.n_variants))
-
         # Custom statistics
         for stat in self.custom_stats:
             kwargs = self.custom_stat_kwargs.get(stat.__name__, {})
             results[stat.__name__] = stat(window, **kwargs)
 
         return results
-
-    def _freq_stat_results(self, matrix, n_variants):
-        """daf_hist (as daf_bin_0 .. daf_bin_{_DAF_N_BINS - 1}) and mu_sfs columns.
-
-        Defers to diversity.mu_sfs / diversity.daf_histogram (the include/exclude
-        scalar reference), so the windowed value equals the scalar over the same
-        variants. An empty window yields 0.0 / an all-zero histogram, matching the
-        scalar on an empty slice and the fused engine's empty-window output.
-        """
-        res = {}
-        n_bins = _DAF_N_BINS
-        if 'mu_sfs' in self.freq_stats:
-            res['mu_sfs'] = (0.0 if n_variants == 0 else
-                             float(diversity.mu_sfs(matrix, missing_data=self.missing_data)))
-        if 'daf_hist' in self.freq_stats:
-            if n_variants == 0:
-                hist = np.zeros(n_bins)
-            else:
-                hist, _ = diversity.daf_histogram(matrix, n_bins=n_bins,
-                                                  missing_data=self.missing_data)
-                hist = np.asarray(hist)
-            for b in range(n_bins):
-                res[f'daf_bin_{b}'] = float(hist[b])
-        return res
 
     @staticmethod
     def _store_result(results: Dict, key: str, val):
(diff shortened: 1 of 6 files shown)


nspope:
The hardcoded number of bins for daf_hist for the streaming path isn't ideal ... but this is a niche statistic and I'm not sure of how to forward arguments correctly from windowed_analysis, so I'm leaving that alone

tests/test_multiallelic.py
```diff
@@ -745,6 +745,130 @@ def test_mu_sfs_and_daf_histogram_run(self, multiallelic_hm):
         assert len(edges) == len(hist) + 1
 
 
+class TestWindowedDafMuSfsMatchesScalar:
+    """0013 Option A: the windowed daf_hist / mu_sfs equal the scalar
```
andrewkern: what's option A? seems like LLMese

tests/test_implementation_parity.py
```diff
@@ -461,3 +466,36 @@ def test_path_matches_scalar(request, condition, stat, path):
...
+
+@pytest.mark.parametrize("condition", list(_CONDITIONS))
+def test_daf_hist_matches_scalar(request, condition):
+    """daf_hist is vector-valued (daf_bin_0 .. daf_bin_{_DAF_N_BINS - 1}), so it
+    sits outside the scalar _STATS matrix. Check its whole-region histogram
+    against diversity.daf_histogram for every engine that computes it: the fused
+    engine (include only) and the WindowedAnalyzer fallback (include and exclude).
+
+    Multi-window / boundary coverage for daf_hist (and mean_nsl) is deferred to the
+    parity-coverage followup (0006-issue-parity-coverage-gaps.md addendum)."""
```
andrewkern: delete this comment as this `.md` that it's referring to must be one of your local docs

tests/test_streaming_kernels.py
```diff
@@ -99,6 +116,45 @@ def test_multiple_stats_equivalent(self, vcz_store, chunk_bp, window_size):
...
+        df_e = windowed_analysis(eager, window_size=5_000, statistics=stats,
+                                 missing_data="include")
+        df_s = windowed_analysis(stream, window_size=5_000, statistics=stats,
+                                 missing_data="include")
+        _assert_frames_equivalent(df_e, df_s)
+
+    @pytest.mark.xfail(reason="streaming + missing_data='exclude' diverges from "
+                              "eager at chunk boundaries for ALL windowed stats "
+                              "(pi included), a pre-existing streaming window-grid "
+                              "issue; see issue-streaming-exclude-window-grid.md",
```
andrewkern: same thing -- this issue doc must be yours locally and not part of the PR. maybe edit this doc to refer to the original issue or this PR?

pg_gpu/windowed_analysis.py
```diff
@@ -2189,29 +2245,51 @@ def windowed_statistics_fused(haplotype_matrix: HaplotypeMatrix,
...
     if 'daf_hist' in statistics:
-        n_daf_bins = 20
-        daf = dac_gpu.astype(cp.float64) / n_hap
-        daf_bin = cp.minimum((daf * n_daf_bins).astype(cp.int32), n_daf_bins - 1)
-        # Composite index: window * n_daf_bins + daf_bin
-        composite = bin_idx * n_daf_bins + daf_bin
-        valid_daf = in_range
-        flat = _scatter_sum(cp.ones_like(composite[valid_daf], dtype=cp.float64),
-                            composite[valid_daf], n_windows * n_daf_bins)
+        from .diversity import _daf_bin_index
```
andrewkern: do we need this import here? isn't it imported up top?

pg_gpu/windowed_analysis.py
```diff
@@ -2122,14 +2168,24 @@ def windowed_statistics_fused(haplotype_matrix: HaplotypeMatrix,
...
+    # (pos == a window end == the next window's start) in the lower window,
+    # disagreeing with n_variants.
+    bin_idx = cp.searchsorted(ws_gpu, positions, side='right') - 1
+    safe = cp.clip(bin_idx, 0, n_windows - 1)
+    in_range = (bin_idx >= 0) & (bin_idx < n_windows) & (positions < we_gpu[safe])
+
+    # Shared per-allele counts (used by daf_hist and mu_sfs). Per-derived-allele,
+    # per-site n_valid -- the same convention as diversity.daf_histogram / mu_sfs,
+    # so the windowed values equal the scalar over each window (include mode).
+    ac_daf = nv_daf = None
```
andrewkern: I noticed that all the tests are using non-overlapping windows, so I tried to see what happens with overlapping windows and there is a pre-existing bug (my bad) that popped out.

` _compute_window_grid` sets `n_per_var = ceil(window_size / step_size)` and maps each variant to up to n_per_var windows — that's what n_variants and the scatter routines consume.

since this PR rewrites the line and adds this comment asserting the equivalence, it'd be good to either narrow the comment to state the overlapping issue or raise a warning/error on step_size < window_size on these stats

pg_gpu/windowed_analysis.py
```diff
@@ -262,10 +276,11 @@ def compute(self, window: WindowData) -> Dict[str, float]:
...
             for stat in self.statistics:
-                if isinstance(stat, str):
-                    results[stat] = np.nan
-                else:
-                    results[stat.__name__] = np.nan
+                name = stat if isinstance(stat, str) else stat.__name__
+                if name in self.FREQ_STATS:
+                    continue  # emitted as their own columns below
+                results[name] = np.nan
+            results.update(self._freq_stat_results(window.matrix, 0))
```
andrewkern: this looks inconsistent-- the comment says fill with NaN, but line 283 fills with zeros. I think the behavior we want is that if a window has no SNPs, we want to fill NaN.

andrewkern (changes requested):
looks generally clean @nspope but I caught a few little issues

tests/test_multiallelic.py
nspope: Whoops, you're right---didn't do my due diligence here cleaning up.

pg_gpu/windowed_analysis.py
nspope: Right, so the freq stats get zeros, not NaNs --- the comment is wrong. This was done for parity with the scalar path. 

But now that I look at it, the behavior for empty windows is inconsistent across stats and engines, and we're not testing for parity with this special case. This is yet another place where the forking dispatch logic makes for inconsistencies, so I think we'll need a specific followup to sort this out.

In the meantime I'll revert this (fill with NaNs) and switch the scalar path to that behavior as well.

nspope: Actually--- I'm just going to revert and drop these stats from the exclude-mode fallback. Prior to this PR, we'd raise an error if either daf_hist or mu_sfs were requested in windowed_analysis with exclude mode. Keeping the various implementations consistent across edge cases is turning into a huge headache, and filling in the gaps is orthogonal to the point of this PR (fixing bugs with multiallelic data). So, I'll log an issue to add the missing mode at some future point.

pg_gpu/windowed_analysis.py
nspope: I've reverted this fix: it's orthogonal to multiallelic stats, and as you point out there there's bigger issues (it needs a separate push)

nspope:
ready for another look @andrewkern, I reverted the partial windowing fix and the exclude-mode implementation, as these are out-of-scope and were pretty cludgey. Both should be dealt with at a later time (especially the windowing, which is a bug not omitted functionality). So the footprint here is just making the already-implemented functions multiallelic-safe.

andrewkern:
this looks ready to merge @nspope -- go for it when ready

PR #162, opened 2026-07-23
Update README for robust use of pip

molpopgen:
`pip install foo` is not robust -- it is possible that the pip in `$PATH` does not correspond to the `python` in `$PATH`.
(This happens a LOT due to venvs, conda, etc., all being mixed together.)
Best practice is to call the pip module from the current interpreter.
See the [tskit](https://tskit.dev/tskit/docs/stable/installation.html#pip) documentation, for example.


nspope:
Thanks!

andrewkern:
Thanks, Kevin!

________________________________

PR #164, opened 2026-07-28
 Unify the biallelic definition across GenotypeMatrix loaders and make biallelic restriction evident with warning

nspope:
#### Background

Two related problems. (1) Several entry points restrict to biallelic sites without telling the user,
so a biallelic-only statistic run on multiallelic data silently returns a partial result. (2) The
loaders disagreed on what "biallelic" even means, and on how to build a dosage -- so the same data
produced different genotype matrices depending on which loader you used, and none of that was
caught by the test suite because nothing compared the loaders.

This PR adds a shared BiallelicOnlyWarning, unifies the biallelic definition and the dosage rule
across all three GenotypeMatrix loaders (from_vcf, from_zarr eager, from_zarr streaming) and
from_haplotype_matrix, and adds a cross-loader parity harness.

#### The unified definition

- Biallelic = at most two distinct *present* alleles (label / reference independent). A {0,2} site,
  or a reference-absent {1,2} site (possible after subsetting), is biallelic. This replaces the
  narrower "codes in {0,1}" notion that wrongly dropped genuinely-biallelic non-{0,1} sites.
- Dosage = count of the chosen alt allele: allele 0 is reference, alt = highest present allele > 0,
  with a sentinel (out of range) when no alt is present so a monomorphic site counts to 0. This
  replaces summing allele indices (wrong for {0,2}) and `gt > 0` (folds distinct alts; breaks on
  reference-absent {1,2}).
- {0,1} input stays byte-identical, including a monomorphic-derived {1}-only site (dosage 2). This
  is what forced "allele 0 = reference" rather than "highest present allele" in the alt choice.
- Both loaders retain monomorphic and all-missing sites (they load faithfully); dropping
  uninformative sites is the job of apply_biallelic_filter, not the loaders. The diploid stats
  tolerate such columns (grm excludes them via its polymorphic mask; pca/ibs guard every division).

The source of truth for biallelic sites is `_biallelic_and_alt(ac)` in genotype_matrix.py.

#### What changed

- _warnings.py: BiallelicOnlyWarning(UserWarning) plus _warn_biallelic_only(n_dropped, *, context,
  stacklevel): coerces the count to int, no-ops at 0, emits a message leading with the context and
  dropped-site count.
- GenotypeMatrix.from_haplotype_matrix: onto the shared definition. Counts the alt allele instead of
  summing indices; kepps {0,2}/{1,2}/monomorphic (previously {0,2}/{1,2} were dropped and index-sum
  gave impossible dosages like 2+2 -> 4); drops >= 3-allele sites + warns. {0,1} bit-identical.
- GenotypeMatrix.from_vcf: moved off scikit-allel's is_biallelic_01() onto the shared definition.
  keeps {0,2}/{1,2}/monomorphic (is_biallelic_01 dropped all three); counts the alt allele; builds
  the allele-count matrix with numpy so scikit-allel's library footprint is now purely VCF parsing
  (read_vcf). Drops >= 3-allele sites + warns.
- GenotypeMatrix.from_zarr (eager AND streaming, via build_genotype_matrix): onto the shared
  definition. Both zarr paths used a raw ploidy-sum with no classification, so a {0,2} site or any
  multiallelic site produced impossible dosages, silently, and disagreed with from_vcf. Now counts
  the alt allele. A >= 3-allele site has no valid 0/1/2 dosage, so its whole row is recoded to -1
  (present but fully missing) -- this preserves the row/chunk alignment the streaming path relies on
  (from_vcf drops the site instead). The warning is surfaced by the callers: the eager path with the
  exact count, the streaming path once per load on the first affected chunk.
- admixture.patterson_d: warn with the count of >2-allele sites it already drops (num=den=0), once
  per top-level D computation (patterson_d / moving_patterson_d / average_patterson_d). D unchanged.
- GenotypeMatrix.apply_biallelic_filter: docstring corrected. It does no allele classification (the
  matrix is biallelic by construction) -- it drops monomorphic (non-segregating) sites. Name kept
  for backwards compatibility; a rename is deferred as a follow-up.

#### Behaviour

- {0,1} input is byte-identical across every loader, silent.
- {0,2} / reference-absent {1,2} / monomorphic sites are now KEPT (and correctly dosed) by all
  loaders; previously from_vcf dropped them and the other paths mis-handled them.
- A >= 3-allele site: from_vcf and from_haplotype_matrix drop it (fewer variants); from_zarr recodes
  its whole row to -1 (present but fully missing). All paths emit one BiallelicOnlyWarning.
- All three GenotypeMatrix loaders now produce byte-identical dosage matrices from equivalent
  biallelic data, and stay consistent on multiallelic data.
- The warning is silenceable by category (from pg_gpu import BiallelicOnlyWarning;
  warnings.filterwarnings("ignore", category=BiallelicOnlyWarning)).

#### Not in scope / deferred

- LD / moments-LD paths keep the old "codes in {0,1}" notion by design (they are on their way to a
  separate multiallelic-correct LD rewrite): haplotype_matrix.apply_biallelic_filter (a moments
  is_biallelic_01 reimplementation), the moments-LD biallelic filter + alt construction, and the
  Rogers-Huff matrix dosage. These three call sites are documented in the LD subproject plan so the
  reconciliation is not lost. rogers_huff itself restricts to biallelic and will use this warning
  there.
- apply_biallelic_filter is misnamed (it drops monomorphic sites, not a biallelic classification).
  Docstring fixed here; a rename or a clearer alias with soft-deprecation is a public-API change
  deferred as a follow-up issue.

#### Verification

- tests/test_biallelic_only_warning.py: the class and helper (subclass, count/context message,
  numpy-scalar count, no-op at 0, silenceable); the shared _biallelic_and_alt definition and alt
  choice; from_vcf keeps {0,2}/{1,2}/monomorphic and drops the triallelic site with correct
  dosages; from_haplotype_matrix bit-identity on {0,1}, alt-code independence, reference-absent
  {1,2}, and the one-missing-haplotype -> fully-missing (-1) case; degenerate-column downstream
  safety (grm and pca_dosage stay finite on explicit monomorphic and all-missing columns);
  patterson_d warns once.
- tests/test_loader_parity.py (new): drives from_vcf, from_zarr (eager), and from_zarr (streaming)
  from a single call_genotype block and asserts they agree byte-for-byte on biallelic data, stay
  consistent on multiallelic data (zarr recodes to a -1 row where from_vcf drops), and that the
  streaming warning fires exactly once across multiple chunks. This is the cross-loader coverage
  that was missing when the three loaders silently diverged.

### DIFF
--- a/pg_gpu/genotype_matrix.py
+++ b/pg_gpu/genotype_matrix.py
@@ -12,6 +12,30 @@ from typing import Optional
 from .accessible import AccessibleMask, resolve_accessible_mask
 
 
+def _biallelic_and_alt(ac):
+    """Per-site biallelic mask and alt-allele index from an (n_var, K) allele-count
+    matrix (any array namespace).
+
+    Biallelic = at most two distinct present alleles. Allele 0 is the reference;
+    ``alt`` is the highest-index present allele > 0, which the 0/1/2 dosage counts --
+    code-independent, so {0,2} and reference-absent {1,2} sites work, while {0,1}
+    stays bit-identical (including a site fixed for allele 1). A site with no alt
+    allele present (all reference / all missing) gets the out-of-range sentinel ``K``
+    so its dosage counts to 0 and never matches the -1 missing code.
+    """
+    xp = cp if isinstance(ac, cp.ndarray) else np
+    present = ac > 0
+    biallelic = present.sum(axis=1) <= 2
+    K = ac.shape[1]
+    if K <= 1:
+        return biallelic, xp.full(ac.shape[0], K)
+    alt_present = present[:, 1:]                                  # alleles 1..K-1
+    has_alt = alt_present.any(axis=1)
+    highest_alt = K - 1 - alt_present[:, ::-1].argmax(axis=1)     # highest present > 0
+    alt = xp.where(has_alt, highest_alt, K)
+    return biallelic, alt
+
+
 class GenotypeMatrix:
     """Diploid genotype matrix with values 0 (hom ref), 1 (het), 2 (hom alt).
 
@@ -278,12 +302,18 @@ class GenotypeMatrix:
         positions = hap_matrix.positions
         xp = cp if isinstance(hap, cp.ndarray) else np
 
-        # A 0/1/2 dosage can only represent {0,1}-coded sites; an allele index
-        # >= 2 would make the paired-haplotype sum a bogus dosage (2+2 -> 4).
-        # Drop those sites (and their positions) and warn with the count, rather
-        # than silently producing garbage. positions shares hap's array
-        # namespace, so the same boolean mask indexes both.
-        biallelic = hap.max(axis=0) <= 1
+        # Biallelic = at most two distinct present alleles; the dosage counts the
+        # alt (highest present allele > 0), so {0,2} and reference-absent {1,2}
+        # sites are handled correctly and {0,1} is unchanged. Sites with >= 3
+        # alleles cannot be a dosage and are dropped with a warning; positions
+        # shares hap's array namespace, so the same mask indexes both.
+        if xp is cp:
+            from ._memutil import allele_counts
+            ac, _ = allele_counts(hap)
+        else:
+            n_a = (int(hap.max()) + 1) if hap.size else 1
+            ac = np.stack([(hap == a).sum(axis=0) for a in range(n_a)], axis=1)
+        biallelic, alt = _biallelic_and_alt(ac)
         n_dropped = int((~biallelic).sum())
         if n_dropped:
             from ._warnings import _warn_biallelic_only
@@ -291,17 +321,18 @@ class GenotypeMatrix:
                 n_dropped, context="GenotypeMatrix.from_haplotype_matrix")
             hap = hap[:, biallelic]
             positions = positions[biallelic]
+            alt = alt[biallelic]
 
         h1 = hap[0::2]  # even indices
         h2 = hap[1::2]  # odd indices
         n_ind = h1.shape[0]
         n_var = h1.shape[1]
 
-        # Chunk over variants to avoid OOM from boolean intermediates
+        # Dosage = per-individual count of the alt allele (0/1/2); a pair with any
+        # missing haplotype -> -1. Chunk over variants to bound memory.
         geno = xp.empty((n_ind, n_var), dtype=xp.int8)
         if xp is cp:
             free_mem = cp.cuda.Device().mem_info[0]
-            # Each variant needs ~4 * n_ind bytes for intermediates
             chunk = max(1, int(free_mem * 0.3 / (n_ind * 4)))
         else:
             chunk = n_var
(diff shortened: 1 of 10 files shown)


pg_gpu/genotype_matrix.py
```diff
@@ -12,16 +12,41 @@
...
     Shape: (n_individuals, n_variants). Missing data encoded as -1.
 
     Biallelic by construction: values are alt-allele dosage (0/1/2), so this
-    structure cannot represent multiallelic genotypes -- ``from_vcf`` filters
-    multiallelic sites out and ``from_haplotype_matrix`` sums paired haplotypes
-    (well-defined only for {0,1} alleles). For multiallelic data use the
-    allele-index ``HaplotypeMatrix``.
+    structure cannot represent multiallelic genotypes. ``from_vcf`` filters
+    multiallelic sites out and ``from_haplotype_matrix`` drops sites with an
+    allele index >= 2 (the dosage is well-defined only for {0,1} alleles); both
```
andrewkern: is this true? isn't this the behavior we are replacing?

tests/test_biallelic_only_warning.py
```diff
@@ -0,0 +1,272 @@
+"""BiallelicOnlyWarning: the shared warning class and its emit helper (0014).
```
andrewkern: what's the (0014)?

pg_gpu/genotype_matrix.py
```diff
@@ -274,18 +299,40 @@ def from_haplotype_matrix(cls, hap_matrix):
...
+        # alt (highest present allele > 0), so {0,2} and reference-absent {1,2}
+        # sites are handled correctly and {0,1} is unchanged. Sites with >= 3
+        # alleles cannot be a dosage and are dropped with a warning; positions
+        # shares hap's array namespace, so the same mask indexes both.
+        if xp is cp:
+            from ._memutil import allele_counts
+            ac, _ = allele_counts(hap)
+        else:
+            n_a = (int(hap.max()) + 1) if hap.size else 1
+            ac = np.stack([(hap == a).sum(axis=0) for a in range(n_a)], axis=1)
```
andrewkern: I'm unclear here-- why are we retaining CPU side ops?

pg_gpu/genotype_matrix.py
```diff
@@ -389,16 +437,26 @@ def from_vcf(cls, path, include_invariant=False, accessible_bed=None,
...
+        # sites are handled correctly and {0,1} is unchanged. Only sites with
+        # >= 3 alleles are dropped (with a warning); monomorphic and all-missing
+        # sites are retained, matching from_haplotype_matrix. The allele-count
+        # matrix is built with numpy -- allel is used only to parse the VCF --
+        # and missing (-1) is never counted, so it matches allel.count_alleles().
+        gt_max = int(gt.max()) if gt.size else -1
+        n_alleles = gt_max + 1 if gt_max >= 0 else 1
+        ac = np.stack([(gt == a).sum(axis=(1, 2)) for a in range(n_alleles)],
+                      axis=1)
+        is_biallelic, alt = _biallelic_and_alt(ac)
```
andrewkern: we still depend on allel, so we could just use count_alleles(), no?

alternatively why not use cupy for this? 

also it strikes me that this is a strange loop-- is there a simple single pass way to do this?

andrewkern (changes requested):
a few notes here for you Nate

pg_gpu/genotype_matrix.py
nspope: You're right this isn't accurate (we now accept arbitrary allele codes so long as only two are segregating)---but the behaviour we're replacing is to sum alleles regardless of coding, which results in nonsensical dosage

nspope:
Ready for another look @andrewkern

andrewkern:
okay looks good @nspope -- merging

Issue #174, opened 2026-08-06
3-population Dz disagrees with moments on the unphased genotype path


andrewkern:
## Summary

On the unphased genotype path (`use_genotypes=True`), `pg_gpu.moments_ld.compute_ld_statistics`
disagrees with `moments.LD.Parsing.compute_ld_statistics` on six `Dz` statistics when there are
**exactly three populations**. Two and four populations agree at ~1e-11, as does the phased
haplotype path at three populations.

Verified against **moments 1.5.3** (so this is not the #258 `pi2` Cython bug, which 1.5.3 fixes).

## Scope

Same VCF handed to both sides, each doing its own loading and filtering -- the form
`tests/test_moments_ld.py` uses. Bins `np.logspace(2, 6, 6)`.

| Populations | Path | Max rel err | Pairs > 1% |
| --- | --- | --- | --- |
| 2 | haplotype | 2.4e-11 | 0 / 75 |
| 2 | genotype | 1.8e-11 | 0 / 75 |
| 3 | haplotype | 2.2e-10 | 0 / 225 |
| **3** | **genotype** | **1.8e-01** | **14 / 225** |
| 4 | haplotype | 3.0e-11 | 0 / 525 |
| 4 | genotype | 2.0e-11 | 0 / 525 |

Affected statistics: `Dz_1_0_1`, `Dz_1_0_2`, `Dz_1_1_2`, `Dz_1_2_2`, `Dz_2_0_2`, `Dz_2_1_2`.
No `DD` or `pi2` pair exceeds 1%, and no `Dz_0_*` deviates -- every affected triple has first
index `ii` in {1, 2}.

## Detail (3 populations, genotype path, rep2)

```
stat         bin       moments        pg_gpu   rel_err   ratio gpu/mom
Dz_1_0_2       1      -18.4762      -21.8291  1.81e-01     +1.181471
Dz_1_0_1       2       91.6808       87.9149  4.11e-02     +0.958923
Dz_2_1_2       1      -12.7526      -12.3986  2.78e-02     +0.972239
Dz_2_1_2       4      -136.496      -133.475  2.21e-02     +0.977863
Dz_2_0_2       1      -131.083      -133.791  2.07e-02     +1.020656
Dz_1_0_2       0      -5.83992      -5.95992  2.05e-02     +1.020548
Dz_1_0_1       3      -293.034      -298.908  2.00e-02     +1.020045
[...]

Issue #148, opened 2026-07-16
Haplotype-to-diploid pairing convention is inconsistent across the package


nspope:
pg_gpu reconstructs diploid individuals from a HaplotypeMatrix in two contradictory ways, and different parts of the package assume different ones. There is no single canonical row order for a HaplotypeMatrix, so for any given matrix at least one consumer pairs the wrong haplotypes together.

The two layouts, for two individuals and one site:

Consecutive: rows are [ind0 hapA, ind0 hapB, ind1 hapA, ind1 hapB]; individual i is rows (2i, 2i+1).

Split-half: rows are [ind0 hapA, ind1 hapA, ind0 hapB, ind1 hapB]; individual i is rows (i, i + n_ind).

#### Who assumes what

Consecutive: GenotypeMatrix.from_haplotype_matrix (hap[0::2], hap[1::2]) and divergence.fst_weir_cockerham (its docstring states "consecutive haplotypes are one diploid individual").

Split-half: relatedness._get_genotype_data (hap[:n_ind], hap[n_ind:]), which feeds ibs and the streaming ibs path; and HaplotypeMatrix._haplotypes_to_gt (the to_zarr writer, gt[...,0] = hap[:n_samples], gt[...,1] = hap[n_samples:]). load_pop_file documents split-half as the intended convention and notes it matches what the zarr genotype source produces. The unused _get_diploid_genotypes also splits split-half.

#### The loaders disagree

from_ts produces consecutive rows. from_ts builds rows as ts.genotype_matrix().T, i.e. tskit sample-node order, and for a diploid simulation nodes 2i and 2i+1 are individual i. On an msprime tree sequence, GenotypeMatrix.from_haplotype_matrix (consecutive) reproduces the tskit per-individual genotypes exactly, while _get_genotype_data (split-half) does not and disagrees with both the truth and the consecutive result.

The zarr writer assumes split-half. Feeding _haplotypes_to_gt a consecutive-layout matrix where individual 0 is 0/0 and individual 1 is 1/1 (rows [0, 0, 1, 1]) yields two heterozygous samples rather than two homozygotes, because it reads the first half as one allele and the second half as the other.

[...]

PR #176, opened 2026-08-06
Cover the unphased genotype estimator in the momentsLD parity tests

andrewkern:
Adds momentsLD parity coverage for the unphased genotype estimator. `tests/test_moments_ld.py`
pinned `use_genotypes=False` at every call site, so the estimator `compute_ld_statistics`
actually defaults to had no end-to-end test against moments.

**Nine of the new tests fail on this branch.** That is intentional -- they pin the moments
contract for two real gaps. No library code is touched here; the fixes are out of scope.

## What's added

- `three_pop_data` / `four_pop_data` parameterized over both estimators (`ids=["hap", "geno"]`),
  passing the same setting to both sides. Adds ~2 s.
- `TestTwoPopGenotypeLD` -- the genotype counterpart of the existing 2-population IM fixtures.
  Marked `slow`: moments' genotype pair loop over that VCF runs ~75 s, and the haplotype
  fixtures already cost about the same.
- `TestAltCodedSiteParity` -- a hand-written VCF carrying one site whose two observed alleles
  are 0 and 2 rather than 0 and 1.
- `TestHaplotypeMatrixGenotypePath` -- the `haplotype_matrix=` entry point under
  `use_genotypes=True`, plus unit checks on the conversion it routes through.

Also cleans up the `.h5` cache moments writes next to each simulated VCF, so a stale cache
cannot shadow a regenerated file.

## Gap 1: which sites the genotype path keeps

`_biallelic_and_alt` calls a site biallelic if at most two alleles are *observed*, regardless
of allele code, taking the highest present allele > 0 as the alt. moments uses scikit-allel's
`is_biallelic_01()`, which requires the observed pair to be exactly `{0, 1}`. A site declaring
`ALT=C,G` where allele 1 is never called is biallelic in the sample but is kept by one side
and dropped by the other.

The broader rule is arguably the better popgen -- the site really is biallelic -- but it means
`moments_ld` is no longer bit-comparable to moments, which is the module's whole purpose. On
`cache/3pop_geno_benchmark_rep2.vcf` it is a one-variant difference (4807 vs 4806), and that
single site is what #174 was actually reporting: it moves `DD` and `pi2`, sums of large
same-sign terms, by well under a percent, while signed `Dz` cancels heavily and moves by whole
percent -- which reads like a broken `Dz` estimator. See
https://github.com/kr-colab/pg_gpu/issues/174#issuecomment-5209573246 for the full trace.

Whether to keep the broad rule and have `moments_ld` opt into `is_biallelic_01`, or narrow it
everywhere, is a design call for this branch. The likely shape is an opt-in flag on
`GenotypeMatrix.from_vcf` / `from_zarr` that `moments_ld` sets, since the alt index is consumed
building dosage and is not retained on the matrix.

Failing: `TestAltCodedSiteParity` (3), `TestTwoPopGenotypeLD` (3 -- the IM VCF carries one such
site). The 3- and 4-population `geno` cases pass, because those simulated fixtures happen not to
contain one. That is exactly why the hand-written VCF is here.

## Gap 2: haplotype pairing (#148)

`compute_ld_statistics(haplotype_matrix=..., use_genotypes=True)` -- the default
`use_genotypes` -- routes through `GenotypeMatrix.from_haplotype_matrix`, which pairs
consecutive rows `(2i, 2i+1)`. The VCF/zarr loaders produce split-half rows `(i, i + n_ind)`,
which is what `load_pop_file` documents and `relatedness._get_genotype_data` reads. So on a
VCF-loaded matrix each "individual" is built from two different people's chromosomes and the
population assignment is scrambled with it:

```
sample_sets['popA']:  from_vcf [0..7]   from_haplotype_matrix [0,1,2,3,8,9,10,11]
64% of genotypes differ
```

That is #148, which already has the full picture -- both conventions, every consumer on each
side, and the loaders (`from_ts` consecutive, VCF/zarr split-half) that make which-one-is-wrong
depend on data provenance. Nothing new is claimed here; this only adds coverage on the
`moments_ld` entry point, which #148 does not mention.

The tests are written to survive whichever order #148 settles on: they load one VCF two ways
and assert the results agree, which has to hold under either canonical order. If #148 lands on
consecutive rows, the assertions stay valid and the loaders move under them.

Failing: `TestHaplotypeMatrixGenotypePath` (3).

## Verification

```
pixi run -e moments pytest tests/test_moments_ld.py -v          # 6 failed, 34 passed, 6 skipped
pixi run -e moments pytest tests/test_moments_ld.py -v --slow -k TwoPopGenotype   # 3 failed, 3 passed
pixi run ruff check pg_gpu/ tests/ examples/                    # clean
```

Every pre-existing test still passes, as do both `hap` and `geno` parameters of the 3- and
4-population fixtures.

### DIFF
--- a/tests/test_moments_ld.py
+++ b/tests/test_moments_ld.py
@@ -1,10 +1,22 @@
 """
 Tests for pg_gpu.moments_ld integration layer.
 
-Validates that pg_gpu produces the same LD and heterozygosity statistics
-as moments for a two-population IM model dataset.
+Validates that pg_gpu produces the same LD and heterozygosity statistics as
+moments, for both estimators: the phased haplotype estimator
+(``use_genotypes=False``) and the unphased genotype estimator
+(``use_genotypes=True``, which is what ``compute_ld_statistics`` defaults to).
+The two are different estimators of the same quantities, so each side must be
+run with the same setting as the other -- a mismatch silently compares apples to
+oranges rather than failing.
 
 Requires the 'moments' pixi environment: pixi run -e moments pytest tests/test_moments_ld.py
+
+Missing data is deliberately not covered on the genotype path. moments builds its
+0/1/2 dosages with ``to_n_alt()`` and no ``fill=`` argument, and scikit-allel
+defaults to ``fill=0``, so a missing call silently becomes homozygous reference;
+pg_gpu drops it from the per-pair sample count instead. The two are not
+comparable once any genotype is missing, so a parity assertion there would be
+pinning the wrong behavior.
 """
 
 import os
@@ -20,6 +32,7 @@ except (ImportError, AttributeError):
     pytest.skip("moments.LD not available (use pixi -e moments)",
                 allow_module_level=True)
 
+from pg_gpu import GenotypeMatrix
 from pg_gpu.moments_ld import (
     compute_ld_statistics,
     _compute_heterozygosity,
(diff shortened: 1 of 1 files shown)


andrewkern:
@nspope this is ready for review

nspope (approved):

PR #180, opened 2026-08-11
LD statistics restricted to biallelic sites

nspope:
pg_gpu's two-locus LD statistics index-inflated on multiallelic-coded input. The matrix/correlation family (`pairwise_r2`, `pairwise_LD_v`, `locate_unlinked`, `zns`, `omega`, `sigma_d2`, windowed `r2`/`ZnS`) treated the raw allele index as a 0/1 indicator, so `hap.T @ hap` inflated on any site coded `{0,2}` or reference-absent `{1,2}`; the moments `DD`/`Dz`/`pi2` tally counted `hap == 1`/`hap == 0` and silently dropped allele-2 haplotypes; and the segregating pre-filters used `sum(max(hap, 0))`, which also inflates.

This PR gives LD a single mode: it restricts to biallelic segregating sites (exactly two present alleles) with arbitrary allele coding, recoding each site's alt allele (the highest present) to a 0/1 indicator so `{0,1}`, `{0,2}`, and reference-absent `{1,2}` all compute correctly.

An earlier iteration additionally matched `moments.LD` by filtering to exactly `{0,1}`-coded sites. That second, coding-specific filter layered on the general one, each behind an `ac_filter` toggle, produced order-dependent bugs. It is removed: there is now one filter, applied unconditionally, everywhere. This intentionally breaks parity with `moments.LD` on inputs containing a two-present-non-`{0,1}` site (e.g. `{0,2}` at a multiallelic record): pg_gpu keeps and recodes it where moments drops it, and the two agree exactly when the input is already `{0,1}`-coded. A site biallelic in the sample should be used regardless of which ALT index its alternate happens to occupy.

#### What changed

The overloaded `apply_biallelic_filter` is replaced by two orthogonal methods: `restrict_to_biallelic` (on `HaplotypeMatrix`, keeps sites with at most two distinct present alleles, allele codes unchanged, drops three-or-more) and `restrict_to_segregating` (on both matrix classes, drops monomorphic sites). Neither recodes. A filter that removes every site returns a valid 0-variant matrix (through the existing `get_subset` empty path) rather than tripping the constructor's non-empty guard, so the reductions fall through to their `m < K -> 0.0` handling instead of raising.

The 0/1 indicator the count kernels require is derived internally by a private `HaplotypeMatrix._biallelic_indicator`: `1` where a haplotype carries the alt (highest present allele), missing preserved as `-1`, identity on `{0,1}` data. It is never exposed as a matrix, so no consumer is handed a relabeled `{1,2}` site whose derived allele masquerades as the reference.

Every LD entry point restricts to biallelic segregating and computes on the indicator, unconditionally. The `ac_filter` toggle is gone from `compute_ld_statistics_gpu_single_pop`/`_two_pops` and the streaming wrappers (they already accumulated on the indicator, so only the toggle is removed). `moments_ld.compute_ld_statistics` loses its `ac_filter` parameter and its `is_biallelic_01` coding filter: the `_moments_is_biallelic_01` helper and the `GenotypeMatrix.from_vcf(biallelic_01_only=...)` option are removed, and `from_vcf` reverts to its two-present-allele default. Its haplotype accumulation now restricts to two-present sites and recodes via the indicator so `{0,2}`/`{1,2}` count correctly rather than relying on a `{0,1}` pre-filter.

Per-variant and per-pair outputs keep their shape and mark out-of-domain entries with a sentinel: `pairwise_r2` and `pairwise_LD_v` return `NaN` for any pair involving a monomorphic or multiallelic site (this also unifies the degenerate `0` to `NaN`, matching `rogers_huff_r`), and `locate_unlinked` returns `False` for multiallelic sites, which start inactive so they never contaminate the windowed `r2`. The scalar reductions restrict-and-shrink internally and emit `BiallelicOnlyWarning` once per top-level call.

A precomputed `r2` matrix passed to `zns`/`omega` may carry `NaN` rows/cols for undefined (monomorphic/multiallelic) sites; a shared `_drop_undefined_sites` shrinks to the defined submatrix before the reduction, mirroring what the object-input path already does, after which `zns`/`omega` run their original reductions verbatim and `_compute_mean_r2` excludes `NaN` cells from its distance mask. A precomputed-matrix input therefore gives the same `zns`/`omega` as the equivalent object input. `windowed_r_squared` tallies its all-pairs counts with the sample-contracting matmul (`X.T @ X` on the indicator, shared with `tally_gpu_haplotypes` via `_tally_pairs_impl`), so peak memory is `O(m^2)`; the missing-data tally uses the same idea on `0/1` indicator planes instead of a per-pair Python loop.

`rogers_huff` is dosage-native: it requires a `GenotypeMatrix` (0/1/2 dosage, biallelic by construction) and rejects a `HaplotypeMatrix`, which a caller converts with `GenotypeMatrix.from_haplotype_matrix`.

#### Behaviour

Segregating biallelic `{0,1}` data is unchanged (bit-identical). Sites coded `{0,2}` or reference-absent `{1,2}` now compute correctly through the indicator instead of index-inflating; they give the same LD as the `{0,1}` relabelling. Monomorphic and three-or-more-allele sites are dropped, with a single `BiallelicOnlyWarning` per call naming the statistic. The one deliberate numeric change on biallelic data is that `pairwise_r2` returns `NaN` rather than `0` for a degenerate (monomorphic) pair, for consistency with `rogers_huff_r`; `zns`/`omega` reduce over the defined pairs, so their scalar output is unaffected. Single-mode LD is loader-invariant: identical whether loaded from a VCF, a tree sequence, or a zarr store. `moments_ld` matches `moments.LD` only on `{0,1}`-coded input.

#### Verification

- `pairwise_r2` equals tskit `ld_matrix` `r2` and `pairwise_LD_v` equals tskit `ld_matrix` `D` to `1e-9` on a binary-mutation tree sequence (the biased plug-in family).
- moments parity: pg_gpu and `moments.LD` agree for every statistic on `{0,1}` data, for both the phased haplotype and unphased genotype estimators, and disagree on a fixture with a two-present-`{0,2}` site; a companion test pins that the plain and alt-coded fixtures are identical except for that site's coding (same alt-dosage matrix, moments dropping only that site), so the divergence is attributable to it alone.
- General-coding equivalence: `{0,2}` and reference-absent `{1,2}` give the same `pairwise_r2`, `zns`, `omega` and windowed `r2` as the `{0,1}` relabelling.
- Loader consistency: single-mode LD is identical across VCF / tree-sequence / zarr on a simulation containing three-or-more-allele and non-`{0,1}` sites.
- `rogers_huff` matches `scikit-allel.rogers_huff_r`.
- Keep-shape convention: `pairwise_r2`/`pairwise_LD_v` are `NaN` on out-of-domain pairs with shape preserved; `locate_unlinked` is `False` on multiallelic sites; `BiallelicOnlyWarning` fires once per top-level call and not at all on all-biallelic input; an all-filtered-away input returns `0.0` rather than raising.

#### Deferred

- Three `test_moments_ld` tests (`TestHaplotypeMatrixGenotypePath`) are xfailed: the issue #148 haplotype-to-diploid pairing inconsistency (`GenotypeMatrix.from_haplotype_matrix` pairs consecutive rows while `from_vcf`/zarr pair per-sample), pre-existing and a separate loader fix.

- Missing-genotype LD statistics are not implemented

- Windowed Rogers-Huff on a `HaplotypeMatrix` (`windowed_r_squared(estimator='rogers_huff')`) requires a manual `GenotypeMatrix.from_haplotype_matrix` conversion, and need a auto-convert follow-up.


pg_gpu/haplotype_matrix.py
andrewkern, line 1671: isn't `self` here a haplotype_matrix? won't this error out now?

pg_gpu/haplotype_matrix.py
```diff
@@ -1747,32 +1791,45 @@ def windowed_r_squared(self, bp_bins, percentile=50, pop=None,
...
         if estimator == 'rogers_huff':
             if pop is not None:
                 raise NotImplementedError(
                     "windowed_r_squared(estimator='rogers_huff', pop=...) "
                     "is not yet implemented; pass the full matrix or "
                     "subset to the desired haplotypes first.")
             from pg_gpu.ld_statistics import _rogers_huff_pairwise_r
+            m = self.num_variants
+            pos = self.positions
             r_full = _rogers_huff_pairwise_r(self)
```
andrewkern, line 1799: same issue -- isn't self the wrong type?

pg_gpu/haplotype_matrix.py
andrewkern: yeah i believe this raises a type error currently

pg_gpu/moments_ld.py
andrewkern, line 178: so this is a problem I think -- if the from_vcf is skipped, ac_filter is skipped and then the tally will be incorrect

pg_gpu/haplotype_matrix.py
```diff
@@ -2117,8 +2174,13 @@ def compute_ld_statistics_gpu_single_pop(
         >>> stats[(0.0, 10000.0)]  # (DD, Dz, pi2) for first bin
         """
         if ac_filter:
-            filtered_self = self.apply_biallelic_filter()
-            return filtered_self.compute_ld_statistics_gpu_single_pop(
+            from ._warnings import _warn_biallelic_only
+            biallelic = self.restrict_to_biallelic()
```
andrewkern: two things about the ac_filter split here (same pattern in the two-pop and streaming versions at 2290, 2421, 2471):

with ac_filter=True we are diverging from momentsLD -- we should probably note this somewhere in docs?

with ac_filter=False we now feed _biallelic_indicator() into the accumulator, which recodes a 3+ allele site as highest-allele-vs-rest. no drop, no NaN, no warning. every other LD path in this PR does one of those three.

i think we need to decide once which rule each entry point uses -- see my comment over in moments_ld.py

pg_gpu/haplotype_matrix.py
```diff
@@ -1090,72 +1090,90 @@ def get_subset_from_range(self, low: int, high: int) -> "HaplotypeMatrix":
...
-        # 2. Both reference (0) and alternate (1) alleles are present
-        # Missing data (-1) is ignored — a site with only 0, 1, and -1 is biallelic
-        is_biallelic = (multiallelic_count == 0) & (ref_count > 0) & (alt_count > 0)
+            chrom_start, chrom_end = self.chrom_start, self.chrom_end
 
-        keep_mask = is_biallelic
-
-        # Get indices of variants to keep
-        keep_indices = xp.where(keep_mask)[0]
+        return HaplotypeMatrix(
```
andrewkern, line 1113: two possible issues here (and the same two in restrict_to_segregating below):

if nothing gets through the filter, we raise "genotypes cannot be empty". zns and omega call this before their own error calls, so a window where every site is multiallelic crashes the whole scan -- the current trunk returned 0.0 for that window.

also the new matrix loses samples and fields. restrict_to_biallelic() then load_pop_file() fails with "No sample names stored", and QC fields loaded via fields= just vanish.

pg_gpu/haplotype_matrix.py
```diff
@@ -1747,32 +1791,45 @@ def windowed_r_squared(self, bp_bins, percentile=50, pop=None,
...
+            biallelic = self.restrict_to_biallelic()
+            _warn_biallelic_only(
+                self.num_variants - biallelic.num_variants,
+                context="windowed_r_squared")
+            m = biallelic.num_variants
+            pos = biallelic.positions
+            ind = biallelic._biallelic_indicator()
+            idx_i, idx_j = cp.triu_indices(m, k=1)
+            pop_idx = biallelic.sample_sets[pop] if pop is not None else None
+            counts_arr, n_valid = compute_counts_for_pairs(
```
andrewkern: this rewrite changes the memory allocs pretty significantly I think?  compute_counts_for_pairs is gonna be quadratic. do we need chunking or some other solution?

also sample_sets[pop] on the line above loses the old check about no sample set defined for a pop name

pg_gpu/genotype_matrix.py
```diff
@@ -451,17 +457,29 @@ def from_vcf(cls, path, include_invariant=False, accessible_bed=None,
         is_biallelic, alt = _biallelic_and_alt(ac)
         _warn_biallelic_only(int(np.sum(~is_biallelic)),
```
andrewkern, line 452: this count runs before biallelic_01_only so the extra sites that are dropped ({0,2}, {1,2}, monomorphic) are removed without being counted. not sure we want this behavior?

pg_gpu/ld_statistics.py
```diff
@@ -798,6 +792,14 @@ def omega(r2_matrix_or_matrix, missing_data='include', estimator='auto'):
     is_hm = isinstance(r2_matrix_or_matrix, HaplotypeMatrix)
     estimator = _resolve_ld_estimator(estimator, is_hm)
 
+    if is_hm:
+        from ._warnings import _warn_biallelic_only
+        biallelic = r2_matrix_or_matrix.restrict_to_biallelic()
```
andrewkern, line 810: omega now filters and warns on every call, and the per-window loop at windowed_analysis.py:2292 calls it once per window. so a scan emits one BiallelicOnlyWarning per window, each with a different count, which will absolutely flood stderr. it also repeats the allele-count pass per window. the fused zns path warns once for the whole scan -- omega should match that.

docs/source/examples.rst
```diff
@@ -144,7 +144,7 @@ Selection Scan Pipeline
    # Get allele counts for binned standardization. Adding up the
    # matrix only counts alternate alleles if your data uses 0 and 1.
    # If a site has a third allele stored as 2, that 2 gets added as a
```
andrewkern: this advice no longer fixes the problem this paragraph describes -- restrict_to_biallelic keeps {0,2} codes, so the dac sum still counts a 2 as 2. the old apply_biallelic_filter guaranteed strict {0,1} and no public method does that now. same problem at missing_data.rst:266. the new tests had to switch to manual strict-{0,1} subsetting for exactly this reason.

pg_gpu/genotype_matrix.py
```diff
@@ -451,17 +457,29 @@ def from_vcf(cls, path, include_invariant=False, accessible_bed=None,
...
+        if biallelic_01_only:
+            # moments' is_biallelic_01: alleles exactly {0, 1}. Stricter than
+            # the <=2-present default -- also drops {0,2} / reference-absent
+            # {1,2} and monomorphic {0}/{1}. On the kept sites the alt is 1,
+            # which _biallelic_and_alt already returns.
+            has_1 = (ac[:, 1] > 0 if ac.shape[1] > 1
+                     else np.zeros(len(ac), bool))
+            no_ge2 = ((ac[:, 2:].sum(axis=1) == 0) if ac.shape[1] > 2
+                      else np.ones(len(ac), bool))
+            keep = (ac[:, 0] > 0) & has_1 & no_ge2
```
andrewkern: i see two problems downstream of this if biallelic_01_only=True:

it conflicts with include_invariant=True -- the mask requires allele 1 present, so all invariant sites are gone before n_total_sites = geno.shape[1] at line 493, and per-site normalization will be wrong with no warning about the conflicting flags.

and if the mask drops every site, pos[0] at line 494 raises IndexError with no hint that the filter emptied the data.

pg_gpu/windowed_analysis.py
andrewkern, line 1: _compute_mean_r2 (line 1416) is not in this diff but it breaks under the new NaN convention: pairwise_r2 now returns NaN for pairs involving a monomorphic or multiallelic site, and the returned mean is not NaN-aware.

andrewkern (changes requested):
okay here are a bunch of comments. sorry this is messy. i was working with claude to run different parts of the code to see what happened and noticed a bunch of weirdness. some of these are decisions we should make up front

pg_gpu/haplotype_matrix.py
nspope: I think that's fine (it's an informative type error) as we'll add a conversion to genotype matrix here in a followup (along with the other genotype matrix stats that can be invoked from haplotype matrix)

pg_gpu/haplotype_matrix.py
nspope: same as above

pg_gpu/moments_ld.py
nspope: This is a tangle, but the issue is:

- We want to support arbitrary allele codes, because that's what we do for the rest of the package (e.g. {0, 2} should not be discarded, as it's biallelic and the allele coding is arbitrary).
- But we also want to reproduce moments.LD exactly with the "drop in replacement"
- And moments.LD restricts to {0, 1} coding only
- So I did this "additional" filtering in the shim only, but this makes for yet more disparity in behaviour depending on the call path.

I'm not quite sure what to do here, will think about it harder. I agree that it's confusing and inconsistent as is.

pg_gpu/windowed_analysis.py
nspope: Good catch

pg_gpu/ld_statistics.py
nspope: Good catch

pg_gpu/haplotype_matrix.py
nspope: I believe it goes from `O(num_variants**2)` to `O(num_haps * num_variants**2)`, but will double check. There's probably some optimization to do here regardless.

nspope: After looking it over there's no need to switch the call to compute_counts_for_pairs, the original tally_gpu_haplotypes should work just fine here (though I may do something to speed up the "missing" case, which reverts to a for loop). The switch was a holdover from an earlier attempt to support general multiallelic data.

andrewkern: okay sounds good

pg_gpu/haplotype_matrix.py
nspope: good catch. it looks like "apply_biallelic_filter" also dropped samples and fields, so (I think) that part is a pre-existing bug. I'll fix it.

nspope: deferred the metadata-dropping to #181

nspope:
This should be good now @andrewkern --- description in the OP is updated.

andrewkern:
okay Nate-- I'll review this today. In reviewing #184, I think that this one should be merged to main first, but if this goes in to the trunk, I think we are good?

nspope:
Yes, either way works (this PR onto trunk; or trunk onto main first and subsequently this PR onto main). The diff should be the same regardless.

andrewkern:
okay this is ready to go-- I'm going to merge this in to trunk and then revisit 184

Issue #159, opened 2026-07-22
Streaming + missing_data='exclude' diverges from eager at chunk boundaries (all windowed stats)


nspope:
Over a StreamingHaplotypeMatrix with missing_data='exclude', windowed_analysis produces a
different per-window result from the eager path: the window start positions and n_variants differ
around chunk boundaries, so every windowed statistic downstream differs there too. It is not
specific to daf_hist / mu_sfs -- plain pi shows it identically. missing_data='include' does not
diverge, and neither does exclude on data with no missing genotypes (exclude then drops nothing).

#### Observed

Fixture: an msprime HaplotypeMatrix with ~10% missing genotypes written to a vcz store, read back
eager and streaming (chunk_bp=25000), window_size=5000, aligned via the usual chrom_start /
chrom_end matching.

- pi, include: window starts and n_variants match eager exactly.
- pi, exclude: window starts and n_variants disagree (e.g. a window starting at 25043 in eager vs
  25031 in streaming), so the per-window pi differs.
- daf_hist / mu_sfs behave the same way: include matches, exclude diverges.

So the divergence is keyed on missing_data='exclude' with missing genotypes actually present, and
it is engine-agnostic (pi runs through the scatter engine, daf_hist / mu_sfs through the fused
engine / fallback), which points at the shared window-grid construction rather than any one stat.

#### Likely mechanism (to confirm)

Under exclude the surviving-variant set differs between the whole-region eager matrix and the
per-chunk streaming matrices, and the window grid appears to re-base per chunk rather than
continuing the global grid, so windows after the first chunk boundary land at different starts.
The exclude filter interacts with how each chunk places its window origin. This needs to be traced
in _stream_windowed_analysis / the per-chunk window construction and reconciled so the streaming
grid is identical to the eager grid regardless of missing_data.

#### Status / scope

Pre-existing: the streaming equivalence tests only ever ran missing_data='include', so streaming +
exclude was never compared to eager. Surfaced while adding streaming coverage for daf_hist /
mu_sfs; the daf_hist / mu_sfs streaming exclude equivalence test is xfailed
(strict=False) against this issue, and the include case passes. Fixing this is a streaming-path
concern independent of the multiallelic work, so it is logged rather than fixed.

andrewkern:
[...]

PR #182, opened 2026-08-14
One window grid and one membership rule across the windowed engines

### LINKED ISSUE #166: Fused windowed engine: boundary variants are assigned to the wrong window by `bin_idx`
#### Boundary membership

In `windowed_statistics_fused`, the per-variant scatter statistics assign each variant to a window with `bin_idx = cp.searchsorted(we_gpu, positions)`, where `we_gpu` is the window end coordinates. This counts how many window ends lie strictly left of a position, so a variant whose position falls exactly on a window boundary (`pos` equal to a window end, which for contiguous windows is also the next window's start) is assigned to the lower window. Everything else in the same engine treats a window as the half-open interval `[start, end)`: `win_start` / `win_stop` come from `searchsorted(positions, [ws, we], side='left')`, so `n_variants = win_stop - win_start` counts a boundary variant in the next window, and the dedicated scatter engines (`_build_scatter_indices`) use `contains = pos < win_stop`, also right-open. So a boundary-coincident variant lands in window `w` for the `bin_idx`-scattered stats but in window `w+1` for `n_variants` and the theta / divergence stats.

Concrete example: positions 0, 100, ..., 11900 (120 sites), windows of 3000 bp with no overlap, so window ends are 3000, 6000, 9000, 12000. The site at `pos == 3000` is reported in `n_variants` for the second window (`[3000, 6000)`), but `bin_idx = searchsorted([3000, 6000, 9000, 12000], 3000) = 0` puts it in the first window's `daf_hist` / `mu_sfs` / `mean_nsl`. So those features for the first window include a variant that `n_variants` (and `pi`, `theta`, `dxy`) attribute to the second.

#### Overlapping windows

`bin_idx = searchsorted(...)` assigns each variant to exactly one window. When `step_size < window_size` a variant belongs to up to `ceil(window_size / step_size)` windows, and `bin_idx` drops it from all but one, so `daf_hist` / `mu_sfs` / `mean_nsl` undercount. The theta / divergence stats do not have this problem: they go through `_compute_window_ranges`, which builds `n_per_var` candidate windows per variant (its docstring: "scatter over (possibly overlapping) windows ... each variant falls inside up to `n_per_var` windows"). So within a single overlapping-window call, `pi` is correct while `daf_hist` / `mu_sfs` are wrong.

Measured: `window_size=2000`, `step_size=1000` over 40 sites at `pos` 0, 100, ..., 3900 gives 3 of 4 windows whose `daf_hist` disagrees with `diversity.daf_histogram` over each window's `[start, end)` slice. It is silent, with no error or warning. It only errors if a non-fusable stat is co-requested, which reroutes to the `WindowedAnalyzer` and raises `Unknown statistic`, and that is incidental.

#### Affected statistics

The per-site `bin_idx` scatter path covers `daf_hist`, `mu_sfs`, and `mean_nsl` (via `_windowed_mean`). The broader `scatter_stats` set (`snp_dist_*`, `mu_var`, `zns`, `omega`, `mu_ld`, `dist_var` / `dist_skew` / `dist_kurt`) should be audited for the same single-window / boundary `bin_idx` pattern when this is picked up. `GenotypeMatrix` is unaffected; this is `HaplotypeMatrix` windowing.

#### Fix direction

Route the `bin_idx` scatter stats through the same window-membership machinery the theta / divergence stats use, `_compute_window_ranges` / `_build_scatter_indices` (`n_per_var` candidate windows per variant, right-open `[start, end)` containment), instead of the single-window `searchsorted`. That fixes both the boundary membership (right-open, matching `n_variants`) and the overlapping-window undercount in one place, and keeps all fused stats on one assignment.

If the full fix is deferred, a stopgap is to reject or warn on overlapping windows (`step_size < window_size`) for these stats, so they stop returning silently wrong values.

#### Validation

For every window, `daf_hist` / `mu_sfs` / `mean_nsl` should equal the scalar computed over that window's own `[start, end)` variant slice, checked across a multi-window grid (not just the whole-region single window the current parity tests use), and across both regimes:

- non-overlapping windows with a variant sitting exactly on an interior boundary (`pos` equal to a window end, which is also the next window's start): it must land in the same window as `n_variants`;
- overlapping windows (`step_size < window_size`): a variant in `k` windows must contribute to all `k`.

The removed multi-window test (`test_multi_window_boundary_matches_scalar`) covered the boundary case for non-overlapping windows and can be restored and extended when the fix lands.

### LINKED ISSUE #179: compute_region anchors windows at the first variant, not at the region start
`WindowedAnalyzer.compute_region(hm, chrom, start, end)` subsets the matrix to
`[start, end)` and then calls `compute`. `WindowIterator._iter_bp_windows` takes its
first window boundary from the data:

```python
chrom_start = int(self.positions_np[0])
chrom_end = int(self.positions_np[-1])
```

It ignores `matrix.chrom_start`, which `get_subset_from_range` has just set to `start`.

On a matrix of 400 variants over 1 Mb, `compute_region(hm, "1", 200000, 300000)` with
20 kb windows returns:

```
 start    end
201251 221251
221251 241251
...
281251 301251
```

The first window starts at 201251, not 200000, and the last one ends at 301251, past the
requested end.

Two effects:

- Windows from neighbouring region calls do not line up, so the results do not
  concatenate into one scan.
- A window can hold variants outside the requested region, or rather, the last window
  reports a span that reaches past it.

This was not visible until #178, because a coordinate-domain bug in the guard made
`compute_region` raise for nearly every real region.

The decision is what `compute_region` promises. Anchoring at `chrom_start` when it is
set, and stopping at `chrom_end`, would make region results tile cleanly. That also
changes `compute` for any matrix whose `chrom_start` sits before its first variant, so
it needs a deliberate choice rather than a quiet patch.

### LINKED ISSUE #208: Mixed single- and two-population statistics crash under missing_data='exclude'
Mixed single- and two-population statistics under
`missing_data='exclude'` crash when missingness is heavy, identically on
main and the multiallelic trunk:

```python
windowed_analysis(hm, window_size=10_000, step_size=10_000,
                  statistics=['pi', 'fst', 'dxy', 'da'],
                  populations=['p1', 'p2'], missing_data='exclude')
# ValueError: Length of values (11) does not match length of index (20)
```

Each request alone works; the mix dies. The dispatcher serves the
single-population statistics from the thetas scatter engine and the
two-population ones from the two-pop scatter engine, then copies the
second frame's columns into the first positionally. Under 'exclude' the
engines drop different sites -- the thetas engine excludes over the
first population, the two-pop engine over both -- and each anchors its
window grid at its own first surviving variant. With 15% whole-genotype
missingness on a JC69 simulation (40+40 haplotypes, 200 kb, 10 kb
windows) that is 20 windows starting at 2,526 in one frame and 11
starting at 39,195 in the other, and the column assignment raises.

This is the window-grid family #182 addresses: with every engine
anchored at the matrix chromosome bounds, both frames tile the same grid
and the merge holds. What #182 does not decide is the semantic half --
whether one returned frame should mix columns computed over two
different exclusion scopes, which deserves a sentence of documentation
either way. Related: #179, #200.

Found by the #184 A/B; reproduced byte-identically on both branches.

andrewkern:
Closes #179. Closes #166. Closes #208.

The two issues sit on different axes of the same module. A windowed engine makes two
decisions: where the window grid starts, and which window each variant belongs to.
#179 was a grid problem in the iterator engine; #166 was a membership problem in the
fused engine. One commit per issue.

## #179: iterator grid anchoring

`WindowIterator._iter_bp_windows` started the grid at the first variant and ignored
`chrom_start` / `chrom_end`, while every fused path anchors at the matrix bounds when
they are set. The engines tiled different grids for the same matrix whenever
`chrom_start` sits before the first variant: `from_ts` (chrom_start 0), a `from_vcf`
region load, or a `get_subset_from_range` subset. `compute_region` was the loudest
symptom -- its windows started at the first variant in the region, so neighbouring
region calls did not tile.

Fix: `_bp_window_bounds` applies the fused rule (matrix bounds when set, variants as
fallback) in both `_iter_bp_windows` and `count_windows`. `count_windows` now counts
the exact candidate starts the loop walks; the old formula under-counted whenever
step and window size differed (progress-bar total only).

## #166: fused per-site window membership

`bin_idx = searchsorted(we_gpu, positions)` assigned each variant to one window, with
a boundary variant (pos == a window end == the next window's start) going to the
lower window. Everything else in the same call -- `n_variants`, the theta and
divergence kernels, the per-window slice stats -- is right-open `[start, end)`. And
one window per variant means overlapping windows (step < window) silently dropped
each variant from every other window covering it. Affected: `daf_hist`, `mu_sfs`,
`mean_nsl`. The audit the issue asked for confirms the rest of the scatter set
(`snp_dist_*`, `mu_var`, `zns`, `omega`, `mu_ld`, `dist_*`) iterates
`win_start[wi]:win_stop[wi]` slices and was already correct.

Fix: membership is the consecutive run `[k_lo, k_hi]` -- `k_hi` the last window whose
start is <= pos, `k_lo` the first whose end is > pos. Both window families the engine
receives (contiguous `bp_bins`, uniform overlapping start/stop arrays) have
nondecreasing starts and ends, so the run is exact. Contiguous grids keep length-one
runs, so the common case keeps its old shape and cost. The chunked engine delegates
these stats to the fused function, so both paths are covered. This restores the
approach of the reverted df9716b and extends it to overlapping windows.

## Tests

Nine new tests; eight fail on main (the ninth pins the unchanged no-bounds fallback):

- per-window `daf_hist` / `mu_sfs` / `mean_nsl` equal the statistic recomputed from
  that window's own `[start, end)` slice, on a boundary-heavy grid (every variant on
  a window edge), on two- and three-deep overlapping grids, and through the direct
  `bp_bins` branch;
- the iterator anchors at `chrom_start`, its grid matches the fused engine, and two
  consecutive `compute_region` calls tile without gaps or overlap.

`pytest tests/ -n 10`: 895 passed, 64 skipped. `ruff`: clean.

## Notes

- The fused engine reports every window; the iterator skips empty ones. Pre-existing
  difference, unchanged here; the cross-engine test compares the non-empty rows.
- A variant exactly at the grid tail (pos == the last window end, when the span is an
  exact multiple of the step) falls outside all windows. Both engines agree on this
  and did before; unchanged.
- This will conflict lightly with the multiallelic branch, which reworks the daf /
  mu_sfs internals next to the membership block (and where df9716b was reverted to
  keep that branch scoped). The membership logic itself is orthogonal to those
  reworks.

## Rebase onto the multiallelic main

The rebase met #184's per-allele rewrite in the three per-site scatter
stats. A reconciliation commit combines both designs: `mean_nsl`,
`daf_hist`, and `mu_sfs` scatter one entry per present (site, allele)
pair -- main's multiallelic conventions (normalized `daf_hist` rows,
`mu_sfs` as edges over segregating entries) -- into every window
membership the shared grid assigns. An adversarial review confirmed the
combination against the scalar functions across 390+ windows and seven
grid geometries, with a first-principles numpy hand audit agreeing
bit-exactly, and found nothing lost from main in the resolution.

## One grid across missing-data modes

The same review showed the unification did not yet cover
`missing_data='exclude'`: the scatter engines computed grid bounds from
the already-filtered matrix, so a mixed request tiled one grid per
statistic -- the #208 crash, and silent wrong-window labels when the
filtered spans happened to agree. Both engines now anchor the grid
before filtering; the trailing window clips at `chrom_end + 1` in the
shared builder (an inclusive base, so the last loaded variant stays in
a partial trailing window and per-base values divide by the bases that
exist); the fused path's variant-domain consumers see the
allele-cap-filtered variant set; and a zero-span grid returns an empty
frame on every path. Twelve regression tests pin the reconciliation and
the grid behavior, each failing before its fix.

Verified at scale on ag1000g X (4.66M variants x 2940 haplotypes):
windowed totals conserve exactly, and the mixed exclude request runs on
the identical grid as include mode.

## Producer-side closure

`exclude_missing_sites` and `filter_variants_by_missing` now keep the
parent chromosome bounds and accessibility mask like the
`restrict_to_*` siblings -- a shared `_keep_sites` helper states the
policy once for all four same-region site filters -- so every
span-normalized exclude-mode statistic divides by the chromosome
rather than the surviving-variant hull, and the grid-anchor defect
class is closed at the producer as well as at the engines. Five tests
pin the preserved context and the denominator; changelog entry
included.

## Remaining, tracked outside this PR

The eager and streaming loaders still choose different bounds
conventions for the same store (the #159 residue; the engines agree
exactly once bounds agree). Unifying them should be decided together
with the region end-convention question in #224, since both define
what bounds a region string implies.

### DIFF
--- a/pg_gpu/windowed_analysis.py
+++ b/pg_gpu/windowed_analysis.py
@@ -2285,9 +2285,37 @@ def windowed_statistics_fused(haplotype_matrix: HaplotypeMatrix,
                                win_start, win_stop, n_windows, statistics,
                                results)
 
-    # Per-site stats binned into windows via scatter_add
-    bin_idx = cp.searchsorted(we_gpu, positions)
-    in_range = (bin_idx >= 0) & (bin_idx < n_windows)
+    # Per-site stats binned into windows via scatter_add. A variant belongs
+    # to window w iff ws[w] <= pos < we[w] -- the same right-open rule
+    # win_start / win_stop and n_variants use. Searching the window ENDS
+    # would put a variant sitting exactly on a boundary (pos == a window end
+    # == the next window's start) in the lower window, disagreeing with
+    # n_variants. Windows may also overlap (step < window), where a variant
+    # belongs to every window covering it, not just one. Both window
+    # families here have nondecreasing starts and ends, so the member
+    # windows of a variant are the consecutive run [k_lo, k_hi]: k_hi is
+    # the last window whose start is <= pos, k_lo the first whose end is
+    # > pos. Row j of site_win holds candidate k_hi - j; site_ok masks the
+    # rows outside the run. For contiguous windows the run has length one
+    # and the arrays stay (1, n_var).
+    if any(s in statistics for s in ('mean_nsl', 'daf_hist', 'mu_sfs')):
+        k_hi = cp.searchsorted(ws_gpu, positions, side='right') - 1
+        k_lo = cp.searchsorted(we_gpu, positions, side='right')
+        if positions.size:
+            n_per_var = max(1, int(cp.max(k_hi - k_lo).get()) + 1)
+        else:
+            n_per_var = 1
+        site_win = (k_hi[None, :]
+                    - cp.arange(n_per_var, dtype=cp.int64)[:, None])
+        site_ok = (site_win >= k_lo[None, :]) & (site_win < n_windows)
+        bin_idx = site_win.ravel()
+        in_range = site_ok.ravel()
+
+        def _per_site(v):
+            """Align a per-variant array with bin_idx / in_range."""
+            if n_per_var == 1:
+                return v
+            return cp.broadcast_to(v, (n_per_var, v.shape[0])).ravel()
 
     # Shared per-allele counts (used by daf_hist and mu_sfs). Per-derived-allele,
     # per-site n_valid -- the same convention as diversity.daf_histogram / mu_sfs,
@@ -2299,21 +2327,10 @@ def windowed_statistics_fused(haplotype_matrix: HaplotypeMatrix,
 
     if 'mean_nsl' in statistics:
         from . import selection as sel
-        # matrix is already population-subsetted upstream; don't re-subset.
-        # nSL is scanned once over this whole (eager) matrix, so each per-site
-        # score sees the full region before being binned into windows. Over a
-        # StreamingHaplotypeMatrix this runs per chunk, truncating the scan at
-        # chunk boundaries (see _stream_windowed_analysis).
-        # nsl() returns (n_var,) for biallelic data and (n_var, n_derived) when
-        # multiallelic focal sites are present; average every finite
-        # (site, allele) score falling in each window.
-        nsl_2d = cp.asarray(sel.nsl(matrix)).reshape(matrix.num_variants, -1)
-        m = nsl_2d.shape[1]
-        vals_flat = nsl_2d.reshape(-1)
-        bin_flat = cp.repeat(bin_idx, m)
-        inrange_flat = cp.repeat(in_range, m)
-        valid = cp.isfinite(vals_flat) & inrange_flat
-        results['mean_nsl'] = _windowed_mean(vals_flat, bin_flat, valid, n_windows)
+        # matrix is already population-subsetted (line 1533); don't re-subset
+        nsl_gpu = _per_site(cp.asarray(sel.nsl(matrix)))
+        valid = cp.isfinite(nsl_gpu) & in_range
+        results['mean_nsl'] = _windowed_mean(nsl_gpu, bin_idx, valid, n_windows)
 
     # SNP distance stats per window
     snp_dist_stats = {'snp_dist_mean', 'snp_dist_var', 'snp_dist_min',
@@ -2359,19 +2376,14 @@ def windowed_statistics_fused(haplotype_matrix: HaplotypeMatrix,
     # allele (freq = count / n_valid), scattered into (window, bin) and
     # normalized per window -- matching diversity.daf_histogram.
     if 'daf_hist' in statistics:
-        n_daf_bins = _DAF_N_BINS
-        derived = ac_daf[:, 1:]                                  # (n_var, K-1)
-        n_der = derived.shape[1]
-        nv = cp.maximum(nv_daf.astype(cp.float64), 1.0)[:, None]
-        freq = (derived.astype(cp.float64) / nv).reshape(-1)     # per (site, allele)
-        # keep present derived alleles at in-range, non-missing sites
-        keep = ((derived > 0) & (nv_daf[:, None] > 0)
-                & in_range[:, None]).reshape(-1)
-        bin_rep = cp.repeat(bin_idx, n_der)                      # site's window
-        composite = cp.where(keep, bin_rep * n_daf_bins
-                             + diversity._daf_bin_index(freq, n_daf_bins), 0)
-        flat = _scatter_sum(keep.astype(cp.float64), composite,
-                            n_windows * n_daf_bins)
+        n_daf_bins = 20
+        daf = dac_gpu.astype(cp.float64) / n_hap
+        daf_bin = cp.minimum((daf * n_daf_bins).astype(cp.int32), n_daf_bins - 1)
+        # Composite index: window * n_daf_bins + daf_bin
+        composite = bin_idx * n_daf_bins + _per_site(daf_bin)
+        valid_daf = in_range
+        flat = _scatter_sum(cp.ones_like(composite[valid_daf], dtype=cp.float64),
+                            composite[valid_daf], n_windows * n_daf_bins)
         hist_matrix = flat.get().reshape(n_windows, n_daf_bins)
         row_sum = hist_matrix.sum(axis=1, keepdims=True)
         hist_matrix = np.divide(hist_matrix, row_sum,
@@ -2383,22 +2395,14 @@ def windowed_statistics_fused(haplotype_matrix: HaplotypeMatrix,
     # muSFS: fraction of segregating (site, allele) entries at the SFS edges
     # (singleton or n_valid-1), per-site n_valid -- matching diversity.mu_sfs.
     if 'mu_sfs' in statistics:
-        derived = ac_daf[:, 1:]
-        n_der = derived.shape[1]
-        nvc = nv_daf[:, None]
-        is_seg = (derived > 0) & (derived < nvc) & in_range[:, None]
-        is_edge = is_seg & ((derived == 1) | (derived == nvc - 1))
-        keep = cp.repeat(in_range, n_der)
-        bin_safe = cp.where(keep, cp.repeat(bin_idx, n_der), 0)
-        seg_sum = _scatter_sum(is_seg.reshape(-1).astype(cp.float64),
-                               bin_safe, n_windows).get()
-        edge_sum = _scatter_sum(is_edge.reshape(-1).astype(cp.float64),
-                                bin_safe, n_windows).get()
-        # matches diversity.mu_sfs, which returns 0.0 when a window has no
-        # segregating (site, allele) entries (rather than NaN)
-        results['mu_sfs'] = np.divide(edge_sum, seg_sum,
-                                      out=np.zeros_like(edge_sum),
-                                      where=seg_sum > 0)
+        is_edge = _per_site(
+            ((dac_gpu == 1) | (dac_gpu == n_hap - 1)).astype(cp.float64))
+        edge_sum = _scatter_sum(is_edge[in_range], bin_idx[in_range], n_windows)
+        total_count = _bin_counts(bin_idx[in_range], n_windows)
+        edge_cpu = edge_sum.get()
+        count_cpu = total_count.get()
+        mu_sfs = np.where(count_cpu > 0, edge_cpu / count_cpu, np.nan)
+        results['mu_sfs'] = mu_sfs
 
     # Per-window pairwise stats (LD, distance moments)
     ld_pairwise = {'zns', 'omega', 'mu_ld'}
--- a/pg_gpu/windowed_analysis.py
+++ b/pg_gpu/windowed_analysis.py
@@ -2294,10 +2294,13 @@ def windowed_statistics_fused(haplotype_matrix: HaplotypeMatrix,
     if any(s in statistics for s in _PER_SITE_SCATTER_STATS):
         k_hi = cp.searchsorted(ws_gpu, positions, side='right') - 1
         k_lo = cp.searchsorted(we_gpu, positions, side='right')
-        if positions.size:
-            n_per_var = max(1, int(cp.max(k_hi - k_lo).get()) + 1)
-        else:
-            n_per_var = 1
+        # n_per_var is grid geometry, not data: coverage depth (windows j
+        # with ws[j] <= x < we[j]) is maximized at a window start, and the
+        # window arrays are already on the host. An upper bound on any
+        # variant's run length; site_ok masks unused rows.
+        depth = (np.searchsorted(ws_cpu, ws_cpu, side='right')
+                 - np.searchsorted(we_cpu, ws_cpu, side='right'))
+        n_per_var = max(1, int(depth.max(initial=0)))
         site_win = (k_hi[None, :]
                     - cp.arange(n_per_var, dtype=cp.int64)[:, None])
         # site_win <= k_hi <= n_windows - 1 by construction, so k_lo is the
--- a/pg_gpu/windowed_analysis.py
+++ b/pg_gpu/windowed_analysis.py
@@ -2315,18 +2315,26 @@ def windowed_statistics_fused(haplotype_matrix: HaplotypeMatrix,
     # Shared per-allele counts (used by daf_hist and mu_sfs). Per-derived-allele,
     # per-site n_valid -- the same convention as diversity.daf_histogram / mu_sfs,
     # so the windowed values equal the scalar over each window (include mode).
-    ac_daf = nv_daf = None
     if any(s in statistics for s in ('daf_hist', 'mu_sfs')):
         from ._memutil import allele_counts
         ac_daf, nv_daf = allele_counts(matrix.haplotypes)   # (n_var, K), (n_var,)
+        derived = ac_daf[:, 1:]                             # (n_var, K-1)
 
     if 'mean_nsl' in statistics:
         from . import selection as sel
-        # matrix is already population-subsetted (line 1533); don't re-subset
-        nsl_gpu = _expand(cp.asarray(sel.nsl(matrix)))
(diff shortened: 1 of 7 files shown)


pg_gpu/windowed_analysis.py
```diff
@@ -1332,19 +1353,11 @@ def windowed_analysis(haplotype_matrix: HaplotypeMatrix,
...
+        chrom_start, chrom_end = _bp_grid_bounds(haplotype_matrix, positions)
 
         # Build window start/stop arrays (supports overlapping windows)
-        win_starts = np.arange(chrom_start, chrom_end, step_size,
-                               dtype=np.float64)
-        win_stops = win_starts + window_size
+        win_starts, win_stops = _bp_window_grid(
+            chrom_start, chrom_end, window_size, step_size)
         # Build equivalent bp_bins for _compute_window_ranges
         bp_bins = np.concatenate([win_starts, [win_stops[-1]]])
```
nspope, line 1391: Very minor edge case, but this'll error out ambiguously on a size-0 grid (which is reachable if chrom_start == chrom_end). Might be worth adding a `if n_windows == 0:` guard.

Here's a MWE:
```python
import numpy as np
from pg_gpu import HaplotypeMatrix, windowed_analysis

pos = np.array([0, 50, 100])
hap = np.tile(np.array([0, 1, 0, 1], dtype=np.int8)[:, None], (1, 3))
hm = HaplotypeMatrix(hap, pos)

# a one-base span should not crash (but currently throws IndexError)
windowed_analysis(hm.get_subset_from_range(50, 51), window_size=100,
                    step_size=100, statistics=['pi'])
```

pg_gpu/windowed_analysis.py
```diff
@@ -672,6 +674,34 @@ def compute_streaming(self, haplotype_matrix: HaplotypeMatrix,
...
+
+
+def _bp_window_grid(chrom_start, chrom_end, window_size, step_size):
+    """Start/stop arrays of the bp window grid.
+
+    Starts run from chrom_start in steps of step_size while strictly below
+    chrom_end; each window spans window_size. Shared by every windowed
+    engine so the grids agree by construction.
+    """
+    win_starts = np.arange(int(chrom_start), int(chrom_end), step_size,
```
nspope, line 703: So I think `chrom_end` is intended to be *inclusive*, but the arange here treats it as exclusive. This allows SNPs to vanish on a window boundary. For example:
```python
import numpy as np
from pg_gpu import HaplotypeMatrix, windowed_analysis

pos = np.array([0, 50, 100])
hap = np.tile(np.array([0, 1, 0, 1], dtype=np.int8)[:, None], (1, 3))
hm = HaplotypeMatrix(hap, pos)

# every variant should land in some window
df = windowed_analysis(hm, window_size=100, step_size=100, statistics=['pi'])
assert df['n_variants'].sum() == hm.num_variants, f"expected {hm.num_variants}, got {df['n_variants'].sum()}"
```

tests/test_windowed_analysis.py
```diff
@@ -978,3 +982,159 @@ def test_matches_allel_per_window(self, seed):
...
+
+    def test_iterator_anchors_at_chrom_start(self):
+        hm = _leading_gap_hm()
+        it = WindowedAnalyzer(window_size=10_000, step_size=10_000,
+                              statistics=['pi'],
+                              progress_bar=False).compute(hm)
+        assert int(it['start'].iloc[0]) == 0
+        assert all(int(s) % 10_000 == 0 for s in it['start'])
+
+    def test_iterator_grid_matches_fused(self):
```
nspope, line 1663: I think it'd be worth: (a) checking multiple stats here, (b) doing overlapping windows

tests/test_windowed_analysis.py
```diff
@@ -978,3 +982,159 @@ def test_matches_allel_per_window(self, seed):
                                         err_msg=f"H123 window {w}")
             np.testing.assert_allclose(row["garud_h2h1"], h2h1_a, rtol=1e-9,
                                         err_msg=f"H2H1 window {w}")
+
+
+class TestFusedPerSiteWindowMembership:
```
nspope, line 1562: This might by a bit orthogonal to the point of this PR, but I think we need a more through two-pop scatter-vs-fused comparison. The existing `test_fused_twopop_missing_matches_scatter` rebuilds windows by hand under a specific missing-data setup. So I think it'd be worth adding some comparisons on an overlapping grid here (i.e. for da/dxy/fst/fst_hudson).

tests/test_windowed_analysis.py
```diff
@@ -978,3 +982,159 @@ def test_matches_allel_per_window(self, seed):
...
+                                   minlength=N_DAF_BINS).astype(float)
+            got_hist = np.array([res[f'daf_bin_{b}'][wi]
+                                 for b in range(N_DAF_BINS)])
+            np.testing.assert_allclose(got_hist, exp_hist)
+            is_edge = (dac[member] == 1) | (dac[member] == self.n_hap - 1)
+            np.testing.assert_allclose(res['mu_sfs'][wi], is_edge.mean(),
+                                       atol=1e-12)
+
+
+class TestEngineGridAgreement:
```
nspope, line 1651: just brainstorming some edge cases to check with tests (feel free to skip):
- single variant windows
- zero-window grid (that's the index error I commented on above)
- that different loaders->windowed_analysis result in the same output

nspope (commented):
Looks great, left a couple comments on edge cases, and spotting one inclusive/exclusive endpoint bug (I think this is a bug, depends how window interval is defined)

andrewkern:
Stress-probing `missing_data='exclude'` for two-population scans turned up a
grid inconsistency, and this branch fixes most of it. Recording what remains
so we can pick it up here.

On trunk-side code, one request family produces three different window grids:

```
include, any request     -> grid 0, 1000, ..., 5000
exclude + fst_wc         -> per-window fallback, grid 15, 1015, ...
exclude without fst_wc   -> scatter engine, grid 166, 1166, ...
```

On this branch the first two collapse into one: the fallback now anchors at
the matrix chromosome bounds through `_bp_grid_bounds`, so `include` and the
fallback-served `exclude` tile the identical 0-anchored grid. That is the
bigger half of the problem gone.

The scatter exclude route still escapes, through a mechanism outside this
diff. It grids off the matrix `exclude_missing_sites` returns, and that goes
through `get_subset`, which builds the new matrix without passing
`chrom_start` / `chrom_end`. The bounds helper then falls back to the first
and last complete variant:

```
exclude without fst_wc   -> starts 101, 1101, ..., 4101   ends ..., 4759
```

Window-by-window against the scalar on the complete-sites matrix: every
window's value is computed over `[start, start + window)` -- membership is
right-open and correct -- but the last window's reported end is clipped to
the final variant. Its value (0.021439) is the scalar over [4101, 5101);
the interval its row claims, [4101, 4759), holds a different site set whose
value is 0.051874. So the number is right and the label is wrong.

Candidate fix: carry the bounds through `get_subset` the way
`get_subset_from_range` already does since #178 -- two lines in its
constructor call. With bounds preserved, the helpers here produce one grid
for every engine and the label clip disappears.

One caveat to weigh first: `get_subset` has many callers, and carrying
bounds changes span-normalization denominators for subset-derived
statistics -- per-base values on a subset would normalize over the parent's
chromosome span instead of the subset's variant span. That may well be the
right semantics, but it moves shipped numbers beyond the windowed engines,
so it deserves its own look rather than a drive-by.

andrewkern:
Two findings from the PR 184 scale stress run land in this PR's territory; recording them here with repros so the grid work covers them.

**Streaming and eager build different window grids for the same region string.** Eager anchors the grid at the first variant position; streaming anchors at the chunk grid (multiples of chunk_bp from origin 0). On ag1000g X:2003707-2600000 (first variant 2003817): eager window starts [2003817, 2103817, ...], streaming [2000000, 2100000, ...] -- a 3817 bp offset, same totals, different per-window membership (on X:2000001-4000000, 6 of 20 windows differ in n_variants; pi maxrel 1.6e-5). The offset is bounded by chunk_bp, so a sparse store or a region opening on a gap can shift windows by hundreds of kb. Both anchors are deliberate per their docstrings, but the same region string returning different per-window values by path is exactly the one-grid-one-membership problem. Repro: scratchpad/audit184/stress/probe_anchor.py and task2_eq.py on the audit machine.

**The streaming chunk grid always starts at origin 0**, so a late region iterates empty chunks proportional to its distance from the origin (X:20000001-22000000 at chunk_bp=500kb builds 44 chunks, needs 4). Results stay correct; the chunk list is just longer than the region. Repro: probe_emptychunks.py.

Neither blocks 184 (both are path-consistency questions, not wrong numbers on either single path), and the second is cosmetic. Flagging so the shared grid design here decides the anchor once for all engines.

andrewkern:
Rebased onto main (post-#184). Correction to my earlier preview comments: the zero-conflict readings were false negatives (the classic merge-tree output prefixes its markers, which the check missed) -- the original #184 sequencing note predicting a conflict here was right. The real overlap was four hunks in windowed_analysis.py, where both lines of work rewrote the per-site scatter stats.

Resolution: the four branch commits replay with their own content, and a fifth commit reconciles the two designs -- mean_nsl, daf_hist, and mu_sfs now scatter one entry per present (site, allele) pair (main's multiallelic convention: normalized daf_hist rows, mu_sfs as edge over segregating entries with 0.0 for empty windows) into every window membership this branch's shared grid assigns (site_win/site_ok, all overlap depths). The membership tests keep their window-matches-own-slice oracle, updated to the normalized histogram; two tests importing _DAF_N_BINS follow the rename.

Full suite on the rebased branch: 1348 passed, 0 failed -- that is main's complete baseline including the #184 windowed-equals-scalar parity guards for these exact statistics, plus this branch's membership tests at every overlap depth. Ruff clean.

andrewkern:
Follow-up to the adversarial review: the engine-side findings are fixed on this branch, so the one-grid claim now holds in exclude mode too.

- The exclude-mode scatter engines anchor their grid before missing-data filtering, so a mixed request tiles one grid: the #208 crash shape runs (verified with the review's own repros, with and without chromosome bounds, including the asymmetric-missing case that crashed), and the silent wrong-label case is gone -- a per-window check confirms every fst value lands on the window its row names.
- The trailing partial window now clips at chrom_end + 1 in the shared grid builder: chrom_end is an inclusive base, so the last loaded variant stays in a partial trailing window, all engines report the same end, and per-base values normalize by the bases that exist (the fused engine previously normalized the trailing window by the full window_size, 3x off on the review's example).
- The fused path's variant-domain consumers see the allele-cap-filtered variant set (a code past the kernel capacity no longer desynchronizes the per-site arrays), and a zero-span grid returns an empty frame instead of raising.

Six regression tests pin these; all fail before the fix. Full suite: 1354 passed. The review's probes re-run green.

Two items remain outside this branch's engine scope, both candidates for follow-up issues: exclude_missing_sites is now the only same-region site filter that drops chromosome bounds (its restrict_* siblings preserve them; aligning it closes this defect class at the producer, but changes scalar span-normalized outputs and needs its own tests), and the eager/streaming loaders still choose different bounds conventions for the same store (the #159 residue -- with bounds forced equal, streaming equals eager exactly).

Issue #173, opened 2026-08-06
Stale committed outputs in pg_gpu_skill_walkthrough.ipynb after the multiallelic pass


andrewkern:
`docs/source/tutorials/pg_gpu_skill_walkthrough.ipynb` ships with stored outputs for 14 of its code cells, including 7 figures. Those outputs were generated before the multiallelic pass, and at least one cell now renders results the current code would not produce.

`conf.py` loads `nbsphinx` but never sets `nbsphinx_execute`, so it uses the default `'auto'`, which skips any notebook that already has outputs. The committed outputs are therefore what readers see on ReadTheDocs, indefinitely, regardless of what the code does now.

### What is stale

**Cell 12 (PCA) -- confirmed stale.** It calls `decomposition.randomized_pca(hm_pca, n_components=10)`. #143 changed that function from a per-variant Patterson standardization to an eigendecomposition of the genetic relatedness matrix over an all-allele one-hot encoding. Both the printed `PC variance explained (top 5)` line and the three-panel PC1/PC2 scatter are values the current implementation would not reproduce.

**Cell 10 (windowed pi) -- needs checking.** It runs `windowed_analysis(hm, statistics=['pi'])` on `X:1-10000000` loaded straight from zarr with no biallelic filter. If that region contains multiallelic sites, the per-allele counting from #127 / #132 changes the windowed pi track. I could not verify this without the source data.

**Probably unaffected.** Cells 14, 16, 18, 20, 22 and 24 all run on `hm_stats`, which is `apply_biallelic_filter()`ed down to strictly `{0,1}` sites, so the diversity, SFS, FST/dxy, iHS and Garud values should be unchanged. Cell 18's SFS now excludes the fixed classes, but a biallelic-filtered matrix has no monomorphic sites, so those bins were already empty. Cell 26 is moments-LD, which #164 deliberately left on the old biallelic definition.

### Why this is not a quick fix

The notebook reads `ag1000g.X.vcf.gz` (~326 GB, per the notebook's own prose) and `ag1000g_metadata.csv`. Neither is in the repo, so re-running it requires whoever has that dataset staged.

### Options

1. Re-run the notebook against the full Ag1000G X data and commit the refreshed outputs. Keeps the tutorial's scale and realism; needs the dataset and a GPU.
[...]

Issue #163, opened 2026-07-24
diversity calculation disagrees with allel and sgkit on human 1000 genomes data


molpopgen:
# Short version

pg_gpu calculates a different value for diversity ("pi") than sgkit and scikit-allel.

The values are:

pg_gpu:  41158.053911185685
sgkit:  41220.16543277111
allel: 41220.165432771115

The difference is about 0.1% which is 10+ orders of magnitude larger than machine precision.
We intentionally disable "span normalize" in order to be able to see the difference.

# Details

## CPU enviroment

The output of `lsb_release -a` is:

No LSB modules are available.
Distributor ID:	Pop
Description:	Pop!_OS 24.04 LTS
Release:	24.04
Codename:	noble

This machine has an AMD 5950x and 128 gigabytes of RAM. 

## NVIDIA environment

The output of `nvidia-smi` is (apologies for the formatting...):

Fri Jul 24 12:12:04 2026       
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 580.159.03             Driver Version: 580.159.03     CUDA Version: 13.0     |
+-----------------------------------------+------------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  NVIDIA TITAN V                 Off |   00000000:0C:00.0 Off |                  N/A |
| 31%   46C    P8             26W /  250W |      23MiB /  12288MiB |      0%      Default |
|                                         |                        |                  N/A |
+-----------------------------------------+------------------------+----------------------+
|   1  NVIDIA TITAN V                 Off |   00000000:0D:00.0  On |                  N/A |
| 29%   43C    P8             27W /  250W |     127MiB /  12288MiB |      2%      Default |
|                                         |                        |                  N/A |
+-----------------------------------------+------------------------+----------------------+

+-----------------------------------------------------------------------------------------+
| Processes:                                                                              |
|  GPU   GI   CI              PID   Type   Process name                        GPU Memory |
|        ID   ID                                                               Usage      |
|=========================================================================================|
|    0   N/A  N/A            1735      G   cosmic-comp                              12MiB |
|    0   N/A  N/A            1812      G   Xwayland                                  8MiB |
[...]

Issue #147, opened 2026-07-16
Streaming genotype population subsetting uses haplotype coordinates


nspope:
Population subsetting on a StreamingGenotypeMatrix selects the wrong rows, because the streaming matrix keeps its sample_sets in haplotype coordinates while a genotype matrix is indexed by individual. Calling a statistic with population= on a StreamingGenotypeMatrix therefore does not restrict to that population; in the observed case `grm(streaming_genotype, population='pop1')` returned the full 18-individual matrix instead of the 9-individual one, because sample_sets['pop1'] held 18 haplotype indices rather than 9 individual indices. The eager path is correct: `GenotypeMatrix.from_zarr (via from_haplotype_matrix)` remaps the haplotype indices to individual indices (i -> i // 2), so `grm(eager_genotype, population='pop1')` restricts correctly.

#### Cause

The eager haplotype-to-genotype conversion remaps sample_sets from haplotype rows to individual rows, but the streaming genotype path carries the source sample_sets through unchanged, so they remain in haplotype coordinates. get_population_matrix (and any per-chunk consumer) then indexes the genotype rows with haplotype indices, which is off by a factor of two and selects the wrong individuals.

#### Fix approach

Remap sample_sets to individual coordinates on the streaming genotype path, the same remap the eager conversion already does, so a StreamingGenotypeMatrix exposes individual-indexed sample_sets. Then population subsetting works identically on the eager and streaming genotype paths. Add a streaming-vs-eager equivalence test for a population-subset statistic on genotype matrices (the existing streaming grm test only exercised haplotype matrices, which is why this was never caught).

#### Scope

This is streaming genotype infrastructure, independent of the multiallelic work in #138; it was surfaced while scoping grm to GenotypeMatrix (grm now rejects haplotype input, so its streaming genotype population path is exercised for the first time). The eager genotype path and all haplotype paths are unaffected.

#### Related

The multiallelic relatedness fix for #138 surfaced it, and should be merged first to avoid conflicts

andrewkern:
Fixed by the sample_sets row-space campaign merged in #184: a genotype stream's sample_sets and materialize(sample_subset=...) speak individual rows end to end, with validation at construction. The release audit's stress run verified the contract at scale: popA and popB resolve to individuals 0-734 and 735-1469 with zero violations across a full chunk walk, and streaming grm/ibs match eager.

PR #184, opened 2026-08-14
Patch statistics for multiallelic data

### LINKED ISSUE #100: diversity.pi gives wrong answer on multi-allelic sites (sum-of-int-values DAC)
## Summary

`diversity.pi` (and the `pi` path through `windowed_analysis`) computes the
derived-allele count per variant as `d = haplotypes.sum(axis=0)`. That works
when haplotype values are strictly `{0, 1}` (biallelic), but on a site whose
haplotype values include a `2` (triallelic under e.g. msprime's default
Jukes-Cantor mutation model), the int sum inflates `d` and the per-variant
pi falls back to a biallelic formula on incorrect counts.

Comparison to scikit-allel's `allel.sequence_diversity` shows ~1% per-window
discrepancy on data containing triallelic sites; on `model="binary"`
msprime data the two libraries agree to FP precision.

## How to reproduce

```python
import msprime
import allel
import cupy as cp
from pg_gpu import HaplotypeMatrix, diversity

# Default mutation model = jc69 -> can produce triallelic sites
ts = msprime.sim_ancestry(samples=20, sequence_length=100_000,
                          recombination_rate=1e-4, random_seed=42, ploidy=2)
ts = msprime.sim_mutations(ts, rate=1e-3, random_seed=42)
hm = HaplotypeMatrix.from_ts(ts)

# pg_gpu
pi_pg = diversity.pi(hm)

# scikit-allel
haps = cp.asnumpy(hm.haplotypes)
pos = cp.asnumpy(hm.positions)
ac = allel.HaplotypeArray(haps.T).count_alleles()
pi_a = allel.sequence_diversity(pos, ac)

print(f"pg_gpu: {pi_pg}, allel: {pi_a}, ratio: {pi_pg / pi_a}")
# observed: pg_gpu and allel disagree by ~1%
```

The same call with `model="binary"` produces matching values to FP precision.

## What's actually wrong

`pg_gpu/diversity.py:_site_contribution`'s `'pi'` branch is:

```python
return cp.where(seg, 2 * d * (n_safe - d) / (n_safe * (n_safe - 1)), 0.0)
```

This assumes `d = count_of_derived_alleles` and `n_safe = total_haps`. With
binary data, `d = haps.sum(axis=0)` is the derived-allele count. With
triallelic data, `haps.sum(axis=0)` mixes alt1 (`+1`) and alt2 (`+2`)
contributions and the resulting `d` no longer corresponds to a single
allele's count.

`allel.mean_pairwise_difference(ac)` handles the general case correctly
because `ac` carries one column per allele and the pairwise sum is over
all allele pairs.

## Why it has been hiding

`tests/test_scikit_allel_comparison.py` constructs haplotype matrices by
hand with values restricted to `{0, 1}`, so `count_alleles()` returns
2-column `ac` and the formulas trivially agree. The discrepancy only
shows up against msprime data with the default JC mutation model, which
came up while building a `test_streaming_vs_allel.py` parity test for
streaming-from-zarr work.

## Suggested fix

Replace the `haps.sum(axis=0)` DAC with a per-allele count from the
underlying `(n_var, n_dip, 2)` genotype block (or extend
`HaplotypeMatrix` to track per-allele counts directly). The pi formula
becomes a sum-over-allele-pairs the same way
`allel.mean_pairwise_difference` handles it. Same change applies to
other diversity / divergence statistics that derive a "derived-allele
count" from the int sum (`theta_h`, `theta_l`, `fay_wu_h`,
`zeng_e`, etc.).

nspope:
Closes #100, #147, #163, and #173. (#147: the streaming genotype row-space fix landed in the sample_sets validation campaign and was verified at scale by the audit's stress run -- zero violations across a full chunk walk.)

This branch grew from "patch statistics for multiallelic data" into the
tree where every statistics fix of the last cycle landed. Merging it
brings main:

* **Multiallelic correctness everywhere.** Alleles are counted
  separately in every estimator -- diversity, divergence, spectra,
  selection, LD -- instead of collapsing alternates into one. The scalar
  and windowed paths agree, and both match scikit-allel and tskit on
  multiallelic and missing data.
* **One haplotype row order.** Sample ``i`` owns rows ``2i`` and
  ``2i + 1`` from every loader, and everything that pairs rows into
  individuals reads that order (#191, #148).
* **Windowed Weir-Cockerham is the diploid per-allele ANOVA** matching
  the scalar and scikit-allel; the old kernel was a haploid estimator
  that duplicated the Hudson column (#188, #133). The chunked engine's
  wrong-row indexing (#189) and the mode-dependent column naming (#193)
  are fixed alongside.
* **sample_sets and population arguments are validated** -- range,
  duplicates, integer type, at assignment and at resolution -- and the
  statistics that pair rows warn on lists that do not pair. Streams
  validate in their own row space, and a genotype stream's whole public
  surface speaks individual rows (#201, #202, #203, #204).
* **The docs walkthrough ships its own data** and runs from a clean
  clone in about four minutes; committed outputs are the published ones
  (#198, #173).
* Plus the moments-popgen pin (#186), tskit-parity ``fst_tskit``,
  ``pca``/``randomized_pca`` on per-allele haplotype columns, and the
  streaming engines' validation and row-space work.

## Verification

Both sides green on this box before comparing: main 886 passed / 64
skipped; trunk (with main merged in) 1314 passed / 78 skipped / 25
xfailed, clean under ``-W error::UnpairedRowsWarning``; ruff clean.

**Value A/B** (identical inputs, one shared environment, three datasets:
clean biallelic msprime, JC69 multiallelic + 15% whole-genotype
missingness, and the shipped Ag1000G subset with real missingness):

* Clean biallelic data: **every statistic identical to 1e-9** across pi,
  thetas, Tajima's D, all four FSTs, dxy, da, spectra, Garud, iHS,
  observed/expected heterozygosity, GRM, IBS, and the windowed
  frames -- except windowed ``fst_wc``, whose main-side sum equals
  main's Hudson column exactly, confirming the haploid duplicate that
  #188 replaced.
* Multiallelic / missing data: 33 differences, every one matching a
  changelog rule, none unexplained.
* **External referee**: on the multiallelic+missing dataset, main
  deviates from scikit-allel (pi 3e-4, Hudson 7e-4, WC 8e-3 relative);
  the trunk matches allel to machine precision (~1e-16) on all three.

**Performance A/B** (same store, region, GPU, and environment; medians
of 3):

| kernel | main | trunk | delta |
| --- | --- | --- | --- |
| load (4.66M var x 2,940 hap) | 130.8 s | 131.5 s | +0.5% |
| pi / theta_w / tajimas_d / sfs | 0.016-0.017 s | 0.085-0.10 s | ~5x rel, ~80 ms abs (was ~17x; #210) |
| dxy / fst_hudson / joint_sfs | 0.21-0.22 s | 0.29-0.30 s | ~1.4x (was 2.3x; #210) |
| fst_wc (scalar) | 0.62 s | 1.48 s | 2.4x (its ANOVA, not counting) |
| windowed single-pop (233 windows) | 0.47 s | 0.37 s | **22% faster** (was +56%; #210) |
| windowed two-pop (fst, fst_wc, dxy, da) | 3.29 s | 3.39 s | +3% |
| ihs | 1082 s | 944 s | **12.8% faster** (was 3079 s; #209) |
| garud_h12 windowed | OOM (109 GB ask) | OOM (109 GB ask) | shared limit |

Single-pass timings for `ihs`; everything else is the median of 3 on an
otherwise idle A100. The two regressions the A/B first measured are
fixed on this branch: `ihs` (was 3079 s) scores every derived allele in
one pair-walk (#209), and the scalar amplification (was ~17x) was a
single line charging a 16-byte global round trip per 1-byte element,
now a shared-memory histogram (#210) -- after which the trunk beats
main on `ihs` and on the windowed single-population engine. What
remains above main is the per-allele arithmetic itself, at absolute
costs no workflow notices. The two-population windowed engine, where
#188 tuned the skip path, is within noise.

## What remains broken, on both sides

One shared pre-existing crash, found by the A/B and reproduced
identically on main and this branch: mixed single- and two-population
statistics under ``missing_data='exclude'`` on heavily missing data.
The thetas scatter engine excludes over the first population and the
two-population engine over both, each anchors its window grid at its
own first surviving variant, and the dispatcher's column merge dies on
the length mismatch (20 vs 11 windows). This is #179/#182 territory --
the shared chrom-anchored grid in #182 removes the crash by
construction -- and needs deciding there, not here.

The A/B findings are filed and the two trunk-side ones are fixed on
this branch: the `ihs` regression (#205 via #209) and the scalar
counting amplification (#206 via #210). The shared, pre-existing Garud
OOM (#207) and mixed-exclude crash (#208) remain open on both sides. Known-open issues that survive
this merge unchanged: #185, #187, #190, #192, #194, #195, #196, #197,
#199, #200.

## Sequencing note

#182 (one window grid across the windowed engines) is based on main and
touches ``windowed_analysis.py``, which this branch rewrites in places.
Merge this branch first: a three-way merge preview shows #182 rebases
onto this trunk with zero textual conflicts, so the order costs nothing,
and #182's suite should run once post-rebase as the semantic check.

## Final audit

A five-way audit ran against this branch at 7de5066: main's full test
suite executed against this code (82 failures, all classified: 63
documented breaks, 18 changelog gaps, 1 real defect), external-referee
numerics on the previously unaudited modules (admixture, decomposition,
ld_statistics, distance_stats, resampling -- all match tskit,
scikit-allel, or numpy oracles to reference precision), an end-to-end
read of the 10.7k-line diff, execution of every documentation snippet
and the walkthrough notebook (bit-identical), and real-data stress at
the 12M-variant / 48 GB scale (streaming and eager agree; auto routing
respects the 50% cap).

Companion PRs from the audit, to land before this merges:
#211 (skill-file drift), #212 (the one real defect: the rogers_huff
estimator path on HaplotypeMatrix), #213 (moments_ld heterozygosity on
the LD site set), #214 (changelog gaps), #215 (filter span bounds and
deterministic diversity_stats order), #216 (dead code and comment
cleanup). Merge mechanics: this trunk contains main, so the merge
commit is exactly this tree.


andrewkern:
A measurement from rebuilding the docs walkthrough on this branch, for reviewers rather than a defect report.

`restrict_to_biallelic` keeps far more sites than it did before this work. Same store, same region (Ag1000G X:1-2,000,000, 1,577,851 records):

| | sites kept |
| --- | --- |
| pre-multiallelic (committed notebook output) | 243,900 (15%) |
| this branch | 1,256,734 (80%) |

The region semantics are unchanged -- both runs report 7,610,313 records for X:1-10,000,000. The source is an all-sites callset where only about 37% of records are polymorphic, so 80% biallelic is the plausible figure and 15% looks too strict. I read this as the intended fix, not a regression.

Worth stating anyway, because it makes downstream work about 5x heavier for the same region. Concretely, the walkthrough's iHS cell went from 244k sites to 1.26M and exhausted an 80 GB A100 on a cohort that previously fit. Anyone with a pipeline tuned to the old counts will want to re-check their memory headroom.

Two related things surfaced while chasing it, both filed separately: #195 (check_diploid_encoding builds ~50 GB of host boolean temporaries per eager load) and #194 (the eager zarr readers copy the genotype block twice).

Issue #187, opened 2026-08-17
Weir-Cockerham has no shared per-site decomposition, so the scatter engine cannot compute fst_wc


andrewkern:
## Background

The windowed engines are kept honest by shared per-site contribution
functions written in CuPy, which several engines consume:

- `diversity._ac_contribution` (`pg_gpu/diversity.py:284`) is used by the
  scalar thetas (`diversity.py:366`), the scatter engine
  (`windowed_analysis.py:824`), and the per-variant engine
  (`windowed_analysis.py:3164`).
- `_twopop_site_components` (`windowed_analysis.py:909`) is used by the
  scatter engine (`1033`) and the per-variant engine (`3044`, `3051`).

Weir-Cockerham has no equivalent. `divergence.fst_weir_cockerham` builds
per-site per-allele `a`, `b`, `c` arrays and collapses them to two floats
inside the same function, so no other engine can reuse the decomposition.

## Consequence

`fst_wc` is the only two-population statistic missing from `scatter_twopop`
(`windowed_analysis.py:1274`, currently `{'fst', 'fst_hudson', 'dxy', 'da'}`)
and the only one with an empty `scatter` slot in the parity registry
(`tests/test_implementation_parity.py:287`). That means:

- A windowed `['fst_wc']` request under `missing_data='include'` has to go
  through the fused CUDA kernel; there is no vectorized CuPy path.
- The parity matrix has no second independent windowed implementation to
  check the kernel against, only the scalar.
- Any future change to the estimator (a ploidy option, r > 2, a correction)
  has to be made twice, once in CuPy and once in CUDA C.

## Suggested change

Split a `_wc_site_components(pop1_haps, pop2_haps, k) -> (a_site, abc_site)`
out of `fst_weir_cockerham`, returning per-site values already summed over
alleles. Then:

- the scalar becomes `sum(a_site) / sum(abc_site)`,
- `_windowed_twopop_scatter` scatter-adds the two vectors per window,
- `'fst_wc'` joins `scatter_twopop` and gains a `scatter` entry in `_STATS`.

That is roughly 30 lines against the ~55 lines of CUDA the kernel carries.

## Out of scope

The CUDA copy should stay. The fused engine exists to answer mixed statistic
requests in a single pass, and the codebase already accepts the same kind of
mirror for Hudson FST and dxy, where the kernel comment names
`divergence._hudson_fst_from_counts` as the source of truth. The gap is the
missing shared CuPy decomposition, not the existence of a second
implementation.

PR #188, opened 2026-08-17
Compute windowed fst_wc with the diploid per-allele ANOVA

### LINKED ISSUE #133: Windowed `fst_wc` disagrees with the scalar `fst_weir_cockerham`
Came up during #132

`windowed_analysis(..., statistics=['fst_wc'], populations=[p1, p2])` routes to the fused CUDA kernel (`_fused_windowed_twopop_kernel`), which computes a **haploid** Weir & Cockerham FST (`h_bar = 0`, each haplotype treated as its own unit). The scalar `divergence.fst_weir_cockerham` is a **diploid** estimator: it pairs consecutive haplotypes into individuals and uses per-allele observed heterozygosity (the `c` component). These are different estimators, so windowed `fst_wc` does not match the scalar `fst_weir_cockerham`.

```
window: fused fst_wc (haploid)   scalar fst_weir_cockerham (diploid)
     0:  -0.000442               -0.000534      (mismatch)
 25000:   0.012111                0.014604      (mismatch)
 50000:   0.015716                0.017199      (mismatch)
```

Reproduce: request `fst_wc` per window via `windowed_statistics_fused(hm, bp_bins=..., statistics=('fst_wc',), pop1=..., pop2=...)` and compare each window to `divergence.fst_weir_cockerham` on the window subset.

If this is intentional, then the function needs to be corrected for multiallelic data; otherwise one of the scalar/windowed paths should be rewritten so they match.

andrewkern:
## Problem

The scalar `fst_weir_cockerham` is a per-allele one-vs-rest ANOVA over diploid
individuals. The fused two-population kernel still used the older form: every
haplotype treated as an independent gamete, all alternate alleles collapsed
into one, and `h_bar` fixed at 0 so the `c` component vanished.

The fused engine serves `missing_data='include'`, which is the default, so
windowed `fst_wc` and the plain function disagreed for most callers. The
per-window Python fallback (`missing_data='exclude'`) matched the scalar, so
which answer you got depended on engine dispatch.

Fixes #133.

The error is a roughly constant **absolute** offset -- at most 0.0012 across 44
datasets spanning true FST 0.0003 to 0.31, median 0.00033. As a fraction that
is under 3% for FST above 0.02 and under 0.1% above 0.15, but it swamps the
near-zero-FST fixtures the parity tests use, where the old kernel ran 67% to
93% low. On biallelic data the old `fst_wc` also agreed with the fused Hudson
FST to ~1e-12, so it was effectively a second copy of the `fst` column.

## Fix

The kernel pairs consecutive haplotypes into individuals, tallies per-allele
copies and per-allele heterozygotes, and runs the same ANOVA as the scalar,
including the `c` term. Both fused entry points share this kernel, so the
chunked engine converts with it.

## Verification

Local GPU runs, since CI has no GPU. Windowed vs scalar, max relative
difference:

| case | fused | chunked |
| --- | --- | --- |
| biallelic | 5.6e-16 | 5.6e-16 |
| multiallelic (JC69, 2597 sites with 3+ alleles) | 4.7e-16 | 4.7e-16 |
| synthetic 6-allele | 1.7e-14 | 1.8e-14 |
| 6-allele + 30% missing | 5.2e-15 | 5.2e-15 |
| half-missing individuals | 2.3e-15 | 2.8e-15 |
| structured A/B split, 15% missing | 2.8e-16 | 2.7e-16 |

Also checked: 1 / 2 / 4 / 100 / 1000 windows (max 6.1e-14), odd-sized
populations, non-contiguous populations, reversed row order, and `from_ts` /
`from_vcf` / `from_zarr` giving identical values.

Both the new kernel and the scalar match scikit-allel's `weir_cockerham_fst`
bit-for-bit on biallelic, multiallelic and odd-count inputs, and on missing
data whenever genotypes are called or missing as whole units. Half-called
genotypes are a convention difference, not a disagreement: allel counts the
lone gamete in its allele frequencies with `n = an // 2`, while pg_gpu's
scalar and both fused engines use complete individuals only -- and all
three pg_gpu paths agree with each other in every case.

Suites after the rebase: 1264 passed, 78 skipped, 25 xfailed (three `fst_wc`
parity cells flip from xfail to pass); moments 39 passed; ruff clean.

## Cost

The kernel does more work per variant, so the ANOVA and the diploid tally are
both skipped when `fst_wc` is not requested. Kernel time against the kernel
this replaces:

| shape | pre-PR | fst_wc not requested | fst_wc requested |
| --- | --- | --- | --- |
| 200 hap x 500k var | 1.17 ms | 1.48 ms | 1.93 ms |
| 800 hap x 300k var | 2.79 ms | 3.14 ms | 3.93 ms |
| 2000 hap x 150k var | 3.58 ms | 3.89 ms | 4.92 ms |

The skip path stays 9% to 26% above the old kernel. That residue is the
register and local-memory footprint of code that is compiled in but not run
(95 registers / 192 B versus 64 / 64 B), which only a separately compiled
variant would remove. `dxy`, `fst_hudson` and `da` are bit-identical with the
skip on or off.

For scale: reaching this kernel without asking for `fst_wc` means pairing a
two-population statistic with a garud/nSL/diploSHIC one, and the kernel is a
few percent of wall time at ordinary window spacing, more when windows overlap
heavily.

## Also in this change

The divergence was recorded in four places and all four are now consistent: the
strict xfail in `test_implementation_parity.py`, the "known exceptions" bullet
in `features.rst`, and the two "(haploid)" annotations in `features.rst` and
`api.rst`. `test_fused_dxy_fst_da_match_scalar` now covers `fst_wc` alongside
`dxy`, `fst` and `da` -- the parity suite compares a single whole-region window,
so that is the only per-window check across many windows.

Follow-up filed as #187: Weir-Cockerham still lacks the shared per-site CuPy
decomposition that `_ac_contribution` and `_twopop_site_components` give the
other statistics, which is why `fst_wc` cannot be computed by the scatter
engine.

## Picked up while landing this

Rebased on the trunk after #191 and #198. The pairing question from review
is settled by #191: every loader now writes sample ``i`` to rows ``2i`` and
``2i + 1`` and ``load_pop_file`` emits that order, so the kernel's
consecutive-row pairing is the package convention, not a one-off. A
pair-preserving shuffle of the population lists changes the result by
exactly zero.

Two bugs in the same dispatch path are fixed here, because windowed
``fst_wc`` cannot be trusted without them:

* #189 -- the chunked engine subsetted the matrix to the single-population
  argument, then indexed that subset with row numbers from the original
  matrix. The dispatcher always passes ``population=pop1`` next to
  ``pop1``/``pop2``, so every chunked two-population scan read wrong
  haplotypes, silently, because CuPy fancy indexing does not bounds-check.
  The rows now come from the original matrix. The old chunked test missed
  this because it never passed ``population=``; the new test mirrors the
  dispatcher's call and fails on the unfixed engine.

* #193 -- the same request named the column ``fst_wc`` under
  ``missing_data='include'`` but ``fst_wc_<p1>_<p2>`` under ``'exclude'``.
  The dispatcher now returns the bare name whenever exactly two
  populations are requested, whichever engine serves it.

## From the adversarial audit

Three review agents (route enumeration, numeric probing, diff review) went
over every path that can emit an FST number. Verdict on the core: sound --
the kernel is a term-for-term transcription of the scalar ANOVA, chunked
equals single-shot to machine precision under adversarial chunking, the
`need_wc` skip is bit-identical for the other statistics, and the streaming
route matches the eager frame exactly. Three surrounding defects they
found, all pre-existing, are fixed here because each can put a wrong or
misleading FST number in front of a user:

* Three or more populations reached the fused engine, which computed the
  first pair only and named the column after the bare statistic -- a
  three-population scan reported `fst` for p1 against p2 and said nothing
  about the rest. Such requests now take the per-window fallback, which
  computes every pair and suffixes the names, matching `'exclude'`.

* The chunked engine looked populations up as `sample_sets` keys, so the
  row-list form the single-shot engine accepts raised `TypeError` -- and
  engine choice depends on free GPU memory, so the same call worked or
  crashed depending on what else was on the card.

* Both fused engines derived the over-cap site filter from the pop1
  subset, so a site with an allele index of 8 or more only in pop2
  survived it and was silently mis-tallied: measured 9.5% off on `fst_wc`
  and 23% on Hudson `fst`, no warning. The cap now reduces over the
  original matrix whenever two-population statistics are requested, so
  the site is dropped and the warning fires. Nucleotide data cannot hit
  this; indel and microsatellite VCFs can.

Known and deliberately left open, all predating this PR: `include` and
`exclude` still disagree on the undefined-FST sentinel (NaN from the fused
engines, 0.0 from the scalar-backed fallback) and on single-population
column names in mixed requests; windowed `da` computes the composite on
the sites both populations have data for, while the scalar `da` lets each
pi term keep its own population's sites, so the two disagree when a site
is entirely missing in one population (the FST flavors and `dxy` are
unaffected); and under `'exclude'` the two serving engines anchor their
window grids differently (#179's family), which the trunk does not yet
have the fix for. Each needs its own decision.

#133, #189 and #193 need closing by hand after merge -- the base branch is
not the default, so the keywords will not fire.

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -372,7 +372,7 @@ Two-population (one kernel launch for all):
 
 - ``fst`` -- Hudson's FST (ratio of averages)
 - ``fst_hudson`` -- alias for ``fst``
-- ``fst_wc`` -- Weir-Cockerham FST (haploid)
+- ``fst_wc`` -- Weir-Cockerham FST
 - ``dxy`` -- absolute divergence
 - ``da`` -- net divergence (Dxy - mean within-pop pi)
 
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -61,6 +61,13 @@ Read this section if you are comparing against older pg_gpu results.
   ``sample_sets`` built by ``load_pop_file`` now lists ``2i`` and
   ``2i + 1`` for each member, and hand-written index lists that assumed
   the previous order need updating.
+* Windowed ``fst_wc`` now gives the same estimate as
+  ``divergence.fst_weir_cockerham``. It used to treat the data as haploid
+  and lump every alternate allele together, so it returned a different
+  number from the plain function on the same variants. Expect windowed
+  values to move: the old ones were off by a small fixed amount, which is
+  under a few percent once FST is above about 0.02 and much larger when
+  FST is near zero. The plain function is unchanged.
 * Frequency spectra: sites where nothing varies no longer contribute,
   folded spectra are returned as ``float64`` rather than integers, and
   ``joint_sfs_folded`` returns a full ``(n1 + 1, n2 + 1)`` grid folded
--- a/docs/source/features.rst
+++ b/docs/source/features.rst
@@ -598,7 +598,7 @@ falls back to a slower route that produces the same numbers.
    * - ``fst``, ``fst_hudson``
      - Hudson's FST per window
    * - ``fst_wc``
-     - Weir-Cockerham FST per window (treats data as haploid; see below)
+     - Weir-Cockerham FST per window
    * - ``dxy``
      - Absolute divergence per window
    * - ``da``
@@ -627,12 +627,9 @@ falls back to a slower route that produces the same numbers.
 
 The rule to expect is that a windowed statistic gives the same answer
 as calling the plain function on just that window's variants. Alleles
-are counted separately here too, the same as everywhere else. There are
-two known exceptions:
+are counted separately here too, the same as everywhere else. There is
+one known exception:
 
-* Windowed ``fst_wc`` treats the data as haploid, so it will not match
-  the plain ``fst_weir_cockerham``, which treats it as diploid. This
-  holds even for two-allele data.
 * When data is missing, the windowed neutrality tests (``tajimas_d``,
   ``normalized_fay_wu_h``, ``zeng_e``, ``zeng_dh``) use the full sample
   size in their variance formula, while the plain versions use an
--- a/docs/source/quickstart.rst
+++ b/docs/source/quickstart.rst
@@ -495,7 +495,7 @@ same values, so your results do not change -- only the speed.
 
 A windowed statistic should give you the same number as calling the
 plain function on just that window's variants. See :doc:`features` for
-the two cases where that is not quite true.
+the one case where that is not quite true.
 
 For advanced usage with custom bin edges, use ``windowed_statistics_fused()``
 directly:
--- a/pg_gpu/windowed_analysis.py
+++ b/pg_gpu/windowed_analysis.py
@@ -1396,7 +1396,21 @@ def windowed_analysis(haplotype_matrix: HaplotypeMatrix,
         span_normalize=span_normalize,
         **kwargs
     )
-    return analyzer.compute(haplotype_matrix)
+    df = analyzer.compute(haplotype_matrix)
+    # The analyzer names two-population columns "<stat>_<pop1>_<pop2>"
+    # because it computes every pair. The fused and scatter engines name
+    # them after the statistic alone when exactly two populations are
+    # given, and the same request must return the same columns whichever
+    # engine serves it. With one pair the bare name is unambiguous, so
+    # rename on the way out; direct analyzer use keeps the suffixed form.
+    if populations is not None and len(populations) == 2:
+        p1, p2 = populations
+        df = df.rename(columns={
+            f"{stat}_{p1}_{p2}": stat
+            for stat in statistics
+            if isinstance(stat, str) and f"{stat}_{p1}_{p2}" in df.columns
+        })
+    return df
 
 
 # Helper functions for built-in statistics
--- a/pg_gpu/windowed_analysis.py
+++ b/pg_gpu/windowed_analysis.py
@@ -1331,7 +1331,14 @@ def windowed_analysis(haplotype_matrix: HaplotypeMatrix,
     requested = set(statistics)
 
     can_fuse = (missing_data == 'include'
-                and requested <= fused_all)
+                and requested <= fused_all
+                # A two-population statistic with anything other than exactly
+                # two populations must not reach the fused engine, which
+                # computes the first pair only and names the column after the
+                # bare statistic. The per-window fallback computes every pair
+                # and suffixes the names, the same as missing_data='exclude'.
+                and (not (requested & fused_two)
+                     or len(populations or []) == 2))
 
     if can_fuse:
         if haplotype_matrix.device == 'CPU':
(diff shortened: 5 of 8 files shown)


nspope:
The haploid estimator seems like the right choice for a haplotype matrix (this was intentional IIRC?) ... I think if it's switched to diploid, then this Fst estimator should become a GenotypeMatrix-scoped function, and go through the HaplotypeMatrix -> GenotypeMatrix conversion route rather than doing it's own haplotype-pairing thing.

andrewkern:
Ah missed that. We can close this if we don’t need it then

________________________________

nspope:
Sorry for being unclear--- I think there's a bug you've found here, so I don't mean to put the kibosh on the PR--- but I think if it's specifically an unphased diploid estimator we should make it a GenotypeMatrix function, so that the haplotype pairing goes through the HaplotypeMatrix -> GenotypeMatrix conversion rather than a one-off.

nspope:
I'm misremembering: #133 says that the haploid/diploid split was pre-existing (and probably unintentional) and I don't think we sorted out which to use. Clearly we should just pick one, you've gone with diploid here. 

So, I guess what I'm suggesting is a follow up which is to scope it to genotype matrix because it's intrinsically diploid---for example, would the estimator be wrong if the data are really haploid---and because there's an existing bug arising from inconsistent haplotype pairing (#148).

Alternatively we could make both implementations haploid which bypasses the ploidy issue entirely.

andrewkern:
yeah agreed the split was accidental, it looks like the kernel was written 4/1 as a copy of the scalar when the scalar was still haploid, then #15 made the scalar diploid and left the kernel behind... ugh
 
on scoping to GenotypeMatrix — i don't think we can? GenotypeMatrix is 0/1/2 dosage so it's biallelic by construction; routing WC  through it drops ~6% of sites on our multiallelic fixture and moves FST by 1%, and we'd lose the exact allel parity we just got. there's also no GenotypeMatrix windowed engine at all, so windowed fst_wc would just go away. and from_haplotype_matrix is still broken cause of #148 -- I think we should probably fix that before anything further -- maybe as part of this PR?

nspope:
Yup, I take the point that everything should be multiallelic as much as possible. I don't think we're tied to allel parity any more (rather tskit, where possible) so I think either haploid or diploid estimator would be fine.

I don't have opinions here, aside from wanting to have all the methods defined on a single data structure (and just convert between data structures where possible). I think if the haplotype-pairing is fixed (to a single infrastructure that's used by everything needing to pair to diploid) then it'll work out fine either way---this seems like as good a place as any to do it.

andrewkern:
i'm punting on this for a minute and instead am working on fixing #148

andrewkern:
#148 is now fixed. moving back to this

PR #191, opened 2026-08-19
Give haplotype rows one order across the package

### LINKED ISSUE #148: Haplotype-to-diploid pairing convention is inconsistent across the package

andrewkern:
Closes #148.

## Problem

A `HaplotypeMatrix` had no single row order. `from_ts` placed a sample's two
gametes adjacent (rows `2i`, `2i+1`); `from_vcf` and `from_zarr` grouped all
of the first gametes ahead of the second ones. Consumers split the same way:

| assumed adjacent rows | assumed the halved layout |
| --- | --- |
| `GenotypeMatrix.from_haplotype_matrix` | `relatedness._get_genotype_data` (`ibs`, `grm`) |
| `divergence.fst_weir_cockerham` | `HaplotypeMatrix._haplotypes_to_gt` (the `to_zarr` writer) |
| `diversity.heterozygosity_observed` | `load_pop_file`'s population index lists |

So for any matrix at least one consumer paired gametes from two different
samples, and which one depended only on where the data came from. Nothing
failed loudly, because a mispaired matrix is still well formed.

## Fix

Adjacent rows are canonical, stated on the `HaplotypeMatrix` docstring.
Producers emit it (`from_vcf`, `build_haplotype_matrix`, `load_pop_file`,
the zarr source's population columns) and the consumers that assumed the
halved layout read it.

Four of the sites are index-space translators rather than array slices, and
are easy to miss when grepping for slicing patterns:

- `zarr_source.slice_subsample` / `slice_subsample_gpu` map a haplotype index
  back to `(sample, ploidy)`. This is the inverse of the population-column
  mapping a few lines away, so leaving it alone would have broken
  `materialize(sample_subset=...)` -- the path `examples/biobank_streaming_scan.py`
  uses.
- `streaming_matrix.materialize` reassembles a subsample into `(n_var, n_dip, 2)`.
  Under the canonical order that split is a free reshape rather than two
  strided writes.
- `HaplotypeMatrix.filter(genotypes=...)` expands a per-sample mask to rows.
- `relatedness._get_haplotype_data`, dead and still carrying the old layout,
  is removed along with the already-dead `_get_diploid_genotypes`.

`from_vcf` now reshapes instead of writing two strided halves: ploidy is the
fastest-varying axis, so merging the trailing axes already interleaves. That
also drops a full copy of the matrix.

## Verification

`TestHaplotypeRowOrder` in `tests/test_loader_parity.py` checks against
outside references rather than pg_gpu against itself:

- `from_ts` and `from_vcf` produce byte-identical matrices, and their
  per-individual genotypes match the tree sequence they came from.
- `ibs` and `grm` match a `GenotypeMatrix` built straight from tskit
  dosages, bypassing haplotype pairing entirely.
- Populations that alternate through the samples, so each one's rows are
  non-contiguous, still pair correctly once subset -- `fst_weir_cockerham`
  pairs consecutive rows *within* the population subset, which is a second
  pairing step layered on the row order.
- A `to_zarr` round trip preserves the order.

Five of the six new cases fail without this change. The three `moments_ld`
cases that were xfailed against this bug now pass, so their marks are gone.

Suites: 1259 passed, 28 xfailed; moments 39 passed with no xfails; ruff clean.

## Results that change

Recorded in the changelog. Statistics that reconstruct individuals from VCF-
or zarr-loaded data -- `ibs`, `grm`, `fst_weir_cockerham`,
`heterozygosity_observed`, `GenotypeMatrix.from_haplotype_matrix`, and
genotype-mode LD -- change value. Statistics that treat rows as independent
gametes are unaffected. Hand-written `sample_sets` index lists that assumed
the previous order need updating.

## Note

Interleaved was the order agreed on the issue. One case is deliberately not
guarded: a hand-written `sample_sets` that separates a sample's two gametes
(e.g. `[0, 4, 1, 5]`) will still mispair, since pairing reads position within
the subset. Every list the library produces is correct; say the word if
`fst_weir_cockerham` should reject that shape instead.

### DIFF
--- a/pg_gpu/relatedness.py
+++ b/pg_gpu/relatedness.py
@@ -532,8 +532,8 @@ def _get_genotype_data(matrix, population=None):
     elif isinstance(matrix, HaplotypeMatrix):
         hap = matrix.haplotypes
         n_ind = hap.shape[0] // 2
-        h1 = hap[:n_ind, :]
-        h2 = hap[n_ind:, :]
+        h1 = hap[0:2 * n_ind:2, :]
+        h2 = hap[1:2 * n_ind:2, :]
         missing = (h1 < 0) | (h2 < 0)
         geno = (cp.maximum(h1, 0) + cp.maximum(h2, 0)).astype(cp.int8)
         geno[missing] = -1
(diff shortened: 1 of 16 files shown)


pg_gpu/relatedness.py
```diff
@@ -532,8 +532,8 @@ def _get_genotype_data(matrix, population=None):
     elif isinstance(matrix, HaplotypeMatrix):
         hap = matrix.haplotypes
         n_ind = hap.shape[0] // 2
-        h1 = hap[:n_ind, :]
-        h2 = hap[n_ind:, :]
+        h1 = hap[0:2 * n_ind:2, :]
+        h2 = hap[1:2 * n_ind:2, :]
         missing = (h1 < 0) | (h2 < 0)
         geno = (cp.maximum(h1, 0) + cp.maximum(h2, 0)).astype(cp.int8)
```
nspope: isn't this `geno = ...` line a bug? Because it is summing allele codes, not "alternative allele"? So this isn't multiallelic safe. I must have missed this in earlier PRs.

Are there any consumers of this function? It seems like it's doing the "genotype matrix" construction, so consumers should just route through a genotype matrix instead---we should only have one path to the dosage matrix.

pg_gpu/streaming_matrix.py
```diff
@@ -606,14 +606,10 @@ def materialize(self, *, region=None, sample_subset=None):
...
-            # = ploidy 1. Assemble the (n_var, n_dip', 2) layout in
-            # cupy so we don't round-trip through a multi-GB host buffer.
-            gt = cp.empty((n_var, n_dip_sub, 2), dtype=gm_gpu.dtype)
-            gt[:, :, 0] = gm_gpu[:, :n_dip_sub]
-            gt[:, :, 1] = gm_gpu[:, n_dip_sub:]
-            del gm_gpu
+            # Canonical row order puts a sample's two gametes adjacent, so the
+            # haplotype axis is already the flattened (n_dip', 2) layout and
+            # the split is a view rather than a copy.
+            gt = gm_gpu.reshape(n_var, n_dip_sub, 2)
```
nspope, line 621: Is there a footgun here? "materialize" has a sample_subset argument, so if the caller asks for a sample subset [0, 2, 4, 6] then will this mistakenly pair 0 and 2?

Maybe just document clearly how these interact, as it's part of the public API?

nspope (approved):
Looks good to me ... spotted one pre-existing multiallelic bug that's adjacent to this PR.

pg_gpu/relatedness.py
andrewkern: good catch. yes this is a bug and i checked and nothing consumes it, i'll delete this

pg_gpu/streaming_matrix.py
andrewkern: oof that would be bad. will add documentation

andrewkern:
okay this is ready. gonna merge to trunk which should now help the haplotype pairing issues

Issue #197, opened 2026-08-22
examples.rst passes per-allele ihs scores into a 1-D standardizer


andrewkern:
`selection.ihs` returns one column per alternate allele (noted in the changelog under Selection). The docs example still treats it as 1-D:

```python
# docs/source/examples.rst:142-148
ihs_raw = selection.ihs(h)
dac = np.sum(h.haplotypes.get() > 0, axis=0)
ihs_std, bins = selection.standardize_by_allele_count(ihs_raw, dac)
```

The `ihs` docstring says `standardize_by_allele_count` takes 1-D input, and that each derived-allele column has to be standardized separately, paired with that allele's per-site derived count. As written the example passes a 2-D array.

This is the same assumption that broke the tutorial notebook, where plotting `ihs` against 1-D positions raised `IndexError: too many indices for array: array is 1-dimensional, but 2 were indexed`. Note that on an all-sites callset the allele indices are sparse, so a site's score can land in any column, not just the first.

`docs/source/index.rst:65` and `.claude/SKILL.md:551-562` may carry the same assumption.

Not reproduced -- reported from reading the example against the docstring.

andrewkern:
Fixed on the trunk by #227: the examples selection pipeline restricts to sites with two alleles in the sample before the scan, so the standardization recipe no longer raises on multiallelic input. Verified against the audit repro (the failing case now completes as a 440-site scan) and the doc-examples harness. Reaches main when #184 merges.

PR #211, opened 2026-08-24
Fix API drift in the skill file and quickstart

andrewkern:
## Summary

A documentation audit of #184 executed every skill-file and quickstart snippet against the trunk. The published docs passed (sphinx build clean, walkthrough notebook reproduces bit-identically). The skill file had drifted off the API on two points that fail on the first executed line, plus four smaller gaps. This fixes all of them.

## Fixes

- `apply_biallelic_filter()` does not exist. All three skill-file sites now call `restrict_to_biallelic()`.
- The selection block and Common Pitfall 5 now teach the multiallelic return shape: `ihs` / `nsl` give `(n_variants,)` on {0,1}-coded data and `(n_variants, K-1)` otherwise, with per-column `standardize_by_allele_count`. The pitfall also notes the filter preserves allele codes, so a surviving `{0,2}` site still produces the 2-D form.
- `heterozygosity_expected` / `heterozygosity_observed` are marked as per-variant arrays in both files; they sat beside scalar stats with no shape note.
- The `sample_sets` example states the haplotype-row contract (sample i owns rows 2i and 2i+1) and the validation behavior: ValueError on out-of-range or duplicate rows, UnpairedRowsWarning when a pairing-aware stat consumes a split row pair.
- `decomposition.pcoa` returns a 2-tuple; the example now unpacks it.
- The quickstart GenotypeMatrix block imports `distance_stats`, which it uses.

## Not touched

The `examples.rst` iHS block still shows the 1-D-only recipe; that is #197 and stays open.

## Verification

All corrected snippets ran on GPU: the 1-D and 2-D selection paths, the {0,2} survival case, heterozygosity shapes, the pcoa unpack, and the three validation behaviors. Ruff clean.

### DIFF
--- a/.claude/SKILL.md
+++ b/.claude/SKILL.md
@@ -194,7 +194,10 @@ Pairwise / cross-window statistics need every variant in scope at the
 same time and **cannot** run chunk-by-chunk: `h.pairwise_r2`,
 `h.locate_unlinked`, `h.windowed_r_squared`'s full r² heatmap form,
 `relatedness.grm`, and `relatedness.ibs`. Calling these on a
-`StreamingHaplotypeMatrix` raises. The escape hatch is `.materialize`:
+`StreamingHaplotypeMatrix` raises. The escape hatch is `.materialize`.
+`grm` / `ibs` take a `GenotypeMatrix` only, so materialize a genotype
+stream (`GenotypeMatrix.from_zarr(..., streaming='always')`) for those
+two -- a materialized haplotype stream still raises `TypeError`:
 
 ```python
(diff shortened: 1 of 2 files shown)


andrewkern:
Added a second commit covering the audit's remaining skill-file findings: the grm/ibs streaming guidance (materialize a genotype stream, not a haplotype stream) and mentions of this cycle's API additions (fst tskit method, patterson_f4, the dosage PCA pair, genetic_relatedness, restrict_to_segregating). Every snippet was executed; the verification caught and fixed two wrong claims before commit (genetic_relatedness takes row-index groups, not pop names; patterson_f4 returns a single per-variant array with a plain-sum scalar).

PR #216, opened 2026-08-24
Post-audit cleanup: dead code, invariant comments, small fixes

andrewkern:
## Summary

The full-diff read of #184 flagged stranded code, comments that narrate project history instead of stating invariants, and a set of small hygiene defects. This applies the survivors.

## Dead code (first commit)

`_site_contribution` and `_prepare_dac` lost their last caller when scalar diversity moved to `_ac_contribution` (the docstring claiming `windowed_analysis` still imports it was false at tip). Transitively dead and removed: `dac_and_n`, its kernel and `chunked_dac_and_n` alias, `chunked_sum_int32`, and `chunked_matmul_accumulate` (no callers on main or trunk). Stale references in the `allele_counts` docstring and SFS weight comment rewritten.

## Comments and docstrings (second commit)

Sixteen sites that referenced replaced code, prior behavior, issue numbers, or deferred plans now state what the code does. The R-parity xfail reason is standalone and `strict=True` (verified: all eight fail deterministically). Three mid-clause docstring wraps fixed. `ihs`/`nsl` document the all-NaN no-alternate-allele return; `moments_ld` no longer claims full argument parity with moments; randomized PCA documents the `n_iter` accuracy trade.

## Code fixes

- Randomized PCA uses a local `cp.random.RandomState` instead of reseeding the process-global generator, so a seeded PCA no longer perturbs later `cp.random` draws. Same-seed outputs change (different draw path), determinism is unchanged.
- `GenotypeMatrix.restrict_to_segregating` forces GPU transfer like every other filter, dropping its numpy branch.
- Dead `n_total_var` assignment in the fused windowed setup removed.
- `debug/stress_test_ag1000g.py` calls current API (`sfs.sfs`, `randomized_pca` without `scaler`); its two pre-existing `allel` lint errors are untouched (out of lint scope).

## Deliberately skipped

- The duplicated fused allele-cap filter (helper + inline copy): refactoring the hot dispatch path is not worth the risk this week; the helper's docstring documents the duplication.
- Function-local `allele_counts` imports (24 sites, 11 files): pure churn, no behavior.
- Promoting `BiallelicOnlyWarning` to a CI error: measured blast radius is 28 failing tests needing per-site judgment; better as its own pass.
- The `.gitignore`-vs-tracked-files question (`.claude/`, `plans/`, `debug/`): policy call, surfaced separately.

## Verification

Full default suite: 1314 passed, 0 failed, 25 xfailed (the 8 parity tests still xfail under strict). Ruff clean on the standard scope.

### DIFF
--- a/debug/stress_test_ag1000g.py
+++ b/debug/stress_test_ag1000g.py
@@ -253,8 +253,8 @@ def main():
     benchmarks.append(("diversity.inbreeding_coefficient",
         lambda: diversity.inbreeding_coefficient(hm, population="pop1"),
         allel_fn(['g'], lambda: np.nanmean(allel.inbreeding_coefficient(_allel_cache['g'].subset(sel1=list(range(N_DIP_PER_POP))))))))
-    benchmarks.append(("diversity.allele_freq_spectrum",
-        lambda: diversity.allele_frequency_spectrum(hm, population="pop1"),
+    benchmarks.append(("sfs.sfs",
+        lambda: sfs.sfs(hm, population="pop1"),
         allel_fn(['g'], lambda: allel.sfs(_allel_cache['g'].count_alleles(subpop=pop1_dip)[:, 1]))))
     benchmarks.append(("diversity.segregating_sites",
         lambda: diversity.segregating_sites(hm, population="pop1"),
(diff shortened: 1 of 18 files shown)


andrewkern:
Third commit from the scale stress run: the zarr loaders' BiallelicOnlyWarning claimed sites were dropped when they are recoded to all-missing rows (measured: 816,082 rows in, 816,082 out, warning said 137,539 dropped), and the streaming message reported one chunk's count as the region's. The wording now matches what each path does and the streaming message names its first-chunk scope. Also: the streaming='never' docstring states the real half-of-free-memory rule, the streaming class documents that each statistic call is a fresh disk pass, and the committed stress harness points at the store's current location and surfaces tracebacks. Full suite 1314 passed.

PR #225, opened 2026-08-24
Pixi task for coverage report and coverage diff

nspope:
Added two pixi tasks to (a) calculate coverage from tests, (b) compare coverage of changed lines relative to some base branch. This is run by, `pixi run coverage && pixi run diff-coverage <branch to compare>`

#### notes

- The localpca tests (which requires R and r2py, missing from the environment) are not run so will be artificially undercovered.
- There's some cludge required to get the environment to use the right libstdc++, this is evidently caused by the kvickio wheel utilizing the system library rather than the environment; currently this is worked around by setting LD_LIBRARY_PATH=... in the pixi coverage task. I think the cludge is fine for the moment as this will only be used by us; but it's noted in the toml and maybe there's a better solution.
-  The coverage run surfaces 26 pre-existing moments.LD parity failures — in test_ld_validation_full.py (17), test_ld_missing_data.py (6), test_ld_statistics_gpu.py (2), and test_ld_validation_synthetic.py (1) — where pg_gpu's GPU computed two-locus statistics (DD/Dz/pi2) diverge from the moments.LD reference. These are unrelated to this PR, which only adds coverage tooling: they reproduce on the base branch itself and are independent of the installed numpy and moments-popgen versions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,6 @@
 
 # pixi environments
 .pixi
-pixi.lock
 *.egg-info
 
 # conda environments
@@ -26,8 +25,6 @@ commit_message.txt
 *.png
 *.csv
 !examples/data/*.csv
-# generated conversion artifact of the tracked im-parsing-example.vcf
-examples/data/*.h5
 
 CLAUDE.md
 TODO_*.md
@@ -47,3 +44,4 @@ htmlcov/
 .coverage
 .coverage.*
 coverage.xml
+diff-coverage.html
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,8 @@ commit_message.txt
 *.png
 *.csv
 !examples/data/*.csv
+# generated conversion artifact of the tracked im-parsing-example.vcf
+examples/data/*.h5
 
 CLAUDE.md
 TODO_*.md
@@ -39,3 +41,9 @@ docs/superpowers/
 # vim for the cool kids
 *.swp
 *.swo
+
+# coverage
+htmlcov/
+.coverage
+.coverage.*
+coverage.xml
--- a/pixi.toml
+++ b/pixi.toml
@@ -16,6 +16,7 @@ msprime = ">=1.0"
 h5py = ">=3.0"
 tqdm = ">=4.0"
 zarr = ">=2.16"
+dask = ">=2024.1"  # pre-2024 dask imports numpy.compat, removed in numpy 2
 pandoc = ">=3.9.0.2,<4"
 nbsphinx = ">=0.9.8,<0.10"
 
@@ -46,6 +47,8 @@ cuda = "12"
 [feature.dev.dependencies]
 pytest = ">=7.0"
 pytest-xdist = ">=3.0"
+pytest-cov = ">=4.0"
+diff-cover = ">=9.5"
 ipython = ">=8.0"
 ipykernel = ">=6.0"
 ruff = ">=0.4"
@@ -75,17 +78,9 @@ sphinx-rtd-theme = ">=1.0"
 python = ">=3.12"
 ruff = ">=0.4"
 
-[feature.cov.dependencies]
-pytest-cov = ">=4.0"
-diff-cover = ">=9.0"
-# Match the default/moments envs; the GPU/moments stack has no 3.14 wheels and a
-# fresh solve would otherwise pick the newest Python.
-python = "3.12.*"
-
 [environments]
 default = ["gpu", "dev", "docs"]
 moments = ["gpu", "dev", "docs", "moments"]
-coverage = ["gpu", "dev", "docs", "moments", "cov"]
 lint = ["lint"]
 
 [tasks]
@@ -94,13 +89,14 @@ test-parallel = { cmd = "pytest tests/ -v -n 10" }
 lint = { cmd = "ruff check pg_gpu/ tests/ examples/" }
 docs = { cmd = "make html", cwd = "docs" }
 
-# FIXME: LD_PRELOAD forces the env libstdc++; the pip kvikio wheel otherwise loads
-# the older system libstdc++, after which importing scipy/moments fails with a
-# missing CXXABI symbol.
-[feature.cov.tasks]
-# Run in order: `pixi run coverage && pixi run diff-coverage <branch to compare>`
-coverage = { cmd = "bash -c 'LD_PRELOAD=$CONDA_PREFIX/lib/libstdc++.so.6 pytest tests/ --slow --cov=pg_gpu --cov-report=term-missing --cov-report=html --cov-report=xml -n 10'" }
+# Coverage tooling runs in the moments env (a superset that also exercises the
+# moments-LD parity tests). Run: `pixi run coverage` then
+# `pixi run diff-coverage <base branch>`.
+[feature.moments.tasks]
+coverage = { cmd = "pytest tests/ --slow --cov --cov-report=term-missing --cov-report=html --cov-report=xml -n 10" }
 
-[feature.cov.tasks.diff-coverage]
-cmd = "diff-cover coverage.xml --compare-branch={{ base }} --format html:htmlcov/diff-coverage.html"
-args = [{ arg = "base", default = "main" }]
+# `base` is substituted unquoted into the shell, so pass a plain git refname;
+# branch names containing shell metacharacters (& ; | ( )) will not work.
+[feature.moments.tasks.diff-coverage]
+cmd = "diff-cover coverage.xml --compare-branch={{ base }} --format html:diff-coverage.html"
+args = [{ arg = "base", default = "origin/main" }]
--- a/pixi.toml
+++ b/pixi.toml
@@ -75,9 +75,17 @@ sphinx-rtd-theme = ">=1.0"
 python = ">=3.12"
 ruff = ">=0.4"
 
+[feature.cov.dependencies]
+pytest-cov = ">=4.0"
+diff-cover = ">=9.0"
+# Match the default/moments envs; the GPU/moments stack has no 3.14 wheels and a
+# fresh solve would otherwise pick the newest Python.
+python = "3.12.*"
+
 [environments]
 default = ["gpu", "dev", "docs"]
 moments = ["gpu", "dev", "docs", "moments"]
+coverage = ["gpu", "dev", "docs", "moments", "cov"]
 lint = ["lint"]
 
 [tasks]
@@ -85,3 +93,14 @@ test = { cmd = "pytest tests/ -v" }
 test-parallel = { cmd = "pytest tests/ -v -n 10" }
 lint = { cmd = "ruff check pg_gpu/ tests/ examples/" }
 docs = { cmd = "make html", cwd = "docs" }
+
+# FIXME: LD_PRELOAD forces the env libstdc++; the pip kvikio wheel otherwise loads
+# the older system libstdc++, after which importing scipy/moments fails with a
+# missing CXXABI symbol.
+[feature.cov.tasks]
+# Run in order: `pixi run coverage && pixi run diff-coverage <branch to compare>`
+coverage = { cmd = "bash -c 'LD_PRELOAD=$CONDA_PREFIX/lib/libstdc++.so.6 pytest tests/ --slow --cov=pg_gpu --cov-report=term-missing --cov-report=html --cov-report=xml -n 10'" }
+
+[feature.cov.tasks.diff-coverage]
+cmd = "diff-cover coverage.xml --compare-branch={{ base }} --format html:htmlcov/diff-coverage.html"
+args = [{ arg = "base", default = "main" }]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -123,21 +123,21 @@ filterwarnings = [
     "ignore::zarr.errors.ZarrDeprecationWarning",
     "ignore:The `compressor` argument is deprecated:UserWarning:zarr",
     # Runtime-visible loader diagnostic; the dedicated tests assert it via local
-    # pytest.warns / simplefilter, which override this global ignore.
-    "ignore::pg_gpu._warnings.BiallelicOnlyWarning",
+    # pytest.warns / simplefilter, which override this global ignore. Matched by
+    # message, not category, so pytest's warnings plugin does not import pg_gpu
+    # before pytest-cov starts measuring (which would drop module-level coverage).
+    "ignore:.* is defined on biallelic sites:UserWarning",
 ]
 
 [tool.coverage.run]
 source = ["pg_gpu"]
-# pythonpath adds pg_gpu/ to sys.path, so modules are importable both as pg_gpu.x
-# and x; relative_files keeps a module attributed to one file path rather than
-# split across the two import names.
+# Record repo-relative file paths in .coverage / coverage.xml instead of
+# absolute ones, so the reports are portable across checkouts and CI (and
+# diff-cover can map the files).
 relative_files = true
-omit = ["tests/*"]
-
-[tool.coverage.report]
-show_missing = true
-skip_covered = false
+# Measure example scripts that tests launch in a child interpreter; pytest-cov 7
+# no longer instruments subprocesses automatically.
+patch = ["subprocess"]
 
 [tool.ruff]
 line-length = 120
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -127,6 +127,18 @@ filterwarnings = [
     "ignore::pg_gpu._warnings.BiallelicOnlyWarning",
 ]
 
+[tool.coverage.run]
+source = ["pg_gpu"]
+# pythonpath adds pg_gpu/ to sys.path, so modules are importable both as pg_gpu.x
+# and x; relative_files keeps a module attributed to one file path rather than
+# split across the two import names.
+relative_files = true
+omit = ["tests/*"]
+
+[tool.coverage.report]
+show_missing = true
+skip_covered = false
+
 [tool.ruff]
 line-length = 120
 target-version = "py312"
--- a/pixi.lock
+++ b/pixi.lock
@@ -1,0 +1,7286 @@
+version: 6
+environments:
+  default:
+    channels:
+    - url: https://conda.anaconda.org/conda-forge/
+    - url: https://conda.anaconda.org/bioconda/
+    indexes:
+    - https://pypi.org/simple
+    options:
+      pypi-prerelease-mode: if-necessary-or-explicit
+    packages:
+      linux-64:
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/_openmp_mutex-4.5-7_kmp_llvm.conda
+      - conda: https://conda.anaconda.org/conda-forge/noarch/_python_abi3_support-1.0-hd8ed1ab_3.conda
+      - conda: https://conda.anaconda.org/conda-forge/noarch/alabaster-1.0.0-pyhd8ed1ab_1.conda
+      - conda: https://conda.anaconda.org/conda-forge/noarch/anyio-4.14.2-pyhcf101f3_0.conda
+      - conda: https://conda.anaconda.org/conda-forge/noarch/apricot-select-0.6.1-pyhd8ed1ab_2.conda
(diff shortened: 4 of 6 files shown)


pixi.toml
```diff
@@ -75,13 +75,32 @@ sphinx-rtd-theme = ">=1.0"
...
 test-parallel = { cmd = "pytest tests/ -v -n 10" }
 lint = { cmd = "ruff check pg_gpu/ tests/ examples/" }
 docs = { cmd = "make html", cwd = "docs" }
+
+# FIXME: LD_PRELOAD forces the env libstdc++; the pip kvikio wheel otherwise loads
+# the older system libstdc++, after which importing scipy/moments fails with a
+# missing CXXABI symbol.
+[feature.cov.tasks]
+# Run in order: `pixi run coverage && pixi run diff-coverage <branch to compare>`
+coverage = { cmd = "bash -c 'LD_PRELOAD=$CONDA_PREFIX/lib/libstdc++.so.6 pytest tests/ --slow --cov=pg_gpu --cov-report=term-missing --cov-report=html --cov-report=xml -n 10'" }
```
andrewkern: **Module-level lines of `pg_gpu` are never measured.**

The pre-existing filter `ignore::pg_gpu._warnings.BiallelicOnlyWarning` in `pyproject.toml:127` makes pytest's warnings plugin import `pg_gpu` (and 21 submodules) inside `pytest_load_initial_conftests`, before pytest-cov starts `Coverage`. Under `-n 10` the xdist workers start measurement later still (`pytest_sessionstart`).

A probe with a tryfirst `pytest_load_initial_conftests` hook (pytest-cov's slot) shows 22 `pg_gpu` modules already in `sys.modules` with the ini as-is, and 0 with the filter replaced. The effect on `pixi run coverage`: about 672 top-level statements (all 48 module-level `cp.RawKernel` constructions included) report as missing on every run, each worker prints "Already imported a file that will be measured" for 22 files, term-missing shows leading `1-40` ranges in every file, and diff-cover marks the `def` line of every new function as uncovered, so diff coverage never reaches 100% for tested code.

Fix: replace line 127 with a message-based filter that needs no import, e.g. `"ignore:.* is defined on biallelic sites:UserWarning"` (`BiallelicOnlyWarning` subclasses `UserWarning`, `pg_gpu/_warnings.py:213`). A `pytest_configure` in conftest is not enough under `-n 10` because workers start coverage after it.

pyproject.toml
```diff
@@ -127,6 +127,18 @@ filterwarnings = [
     "ignore::pg_gpu._warnings.BiallelicOnlyWarning",
 ]
 
+[tool.coverage.run]
```
andrewkern, line 132: **Subprocess coverage is lost with pytest-cov 7.**

`pytest-cov = ">=4.0"` resolves to 7.x, which removed automatic subprocess measurement ("This was removed in pytest-cov 7 - use coverage's patch options"). Three tests run example scripts in a child interpreter via `subprocess.run([sys.executable, ...])`: `tests/test_examples_run.py:22` (`examples/scikit_allel_comparison.py`) and `tests/test_biobank_streaming_example.py:65,113` (`examples/biobank_streaming_scan.py`). With pytest-cov 7.1.0 (no `.pth`, no `COV_CORE_*`) and no `COVERAGE_PROCESS_CONFIG`, the child writes no `.coverage.*` file, so the streaming/VCZ scan paths, `sigma_d2_geno`, etc. reached only through those scripts show as missing in `coverage.xml`, `htmlcov`, and diff-cover.

Fix: add `patch = ["subprocess"]` under `[tool.coverage.run]` (needs coverage >= 7.10).

pixi.toml
andrewkern: **The `bash -c '...'` wrapper silently discards appended arguments.**

pixi places extra args after the closing quote, where bash binds them to `$0`/`$@` and pytest never sees them. `pixi run coverage tests/test_sfs.py -k pi` renders `bash -c '...-n 10' tests/test_sfs.py -k pi` (verified with `--dry-run`); bash sets `$0=tests/test_sfs.py`, `$@=-k pi`, and the full `--slow -n 10` GPU suite runs with no warning while the developer believes a subset ran. `-x`, `--lf`, and `pixi run coverage report`/`html` (intending coverage.py's CLI, which this task name shadows) behave the same way. The existing `test` and `test-parallel` tasks forward args correctly.

pixi's task shell accepts `LD_PRELOAD=$CONDA_PREFIX/lib/libstdc++.so.6 pytest ...` inline (verified), and tasks also have an `env = {}` field, so the wrapper can go.

pixi.toml
```diff
@@ -75,13 +75,32 @@ sphinx-rtd-theme = ">=1.0"
...
 test = { cmd = "pytest tests/ -v" }
 test-parallel = { cmd = "pytest tests/ -v -n 10" }
 lint = { cmd = "ruff check pg_gpu/ tests/ examples/" }
 docs = { cmd = "make html", cwd = "docs" }
+
+# FIXME: LD_PRELOAD forces the env libstdc++; the pip kvikio wheel otherwise loads
+# the older system libstdc++, after which importing scipy/moments fails with a
+# missing CXXABI symbol.
+[feature.cov.tasks]
+# Run in order: `pixi run coverage && pixi run diff-coverage <branch to compare>`
```
andrewkern: **The documented `pixi run coverage && pixi run diff-coverage <branch>` never reaches step two while any test fails.**

pytest exits 1, `bash -c` and pixi forward that status (verified: `pixi run ... bash -c 'exit 3'` -> 3), and `&&` short-circuits, even though `coverage.xml` and `htmlcov/` were already written. The PR body itself reports 26 failing moments.LD parity tests in this exact run (they reproduce on `main`), so as documented the second step is never executed. A developer waits for the full `--slow` GPU run and gets no diff report, with no hint that step two was skipped.

Use `;`, or document the two commands separately. A pixi `depends-on` chain would also stop on failure.

pixi.toml
```diff
@@ -75,13 +75,32 @@ sphinx-rtd-theme = ">=1.0"
 python = ">=3.12"
 ruff = ">=0.4"
 
+[feature.cov.dependencies]
+pytest-cov = ">=4.0"
+diff-cover = ">=9.0"
+# Match the default/moments envs; the GPU/moments stack has no 3.14 wheels and a
+# fresh solve would otherwise pick the newest Python.
+python = "3.12.*"
```
andrewkern: **The `python = "3.12.*"` pin and its comment are wrong.**

The default and moments envs solve to Python 3.14.5 with cp314 `cupy_cuda12x-14.1.1` and `moments_popgen-1.5.3` wheels (published Feb-Jun 2026, before this PR's commit). So the pin does not match those envs; it creates a third, differently-solved GPU stack (cp312 wheels, py312 conda builds) that `pixi run test` and `pixi run -e moments` never exercise. Results from `pixi run test` and `pixi run coverage` can differ, and any difference gets written off as "pre-existing".

One concrete divergence: the tests insert the gitignored repo-root `moments/` checkout onto `sys.path`, and its stale cp312 `.so` files (built 2025-02-16, before the pi2 sign fix) import under 3.12 but not under 3.14, so the coverage env silently defeats the `moments-popgen >= 1.5.3` pin for genotype-mode comparisons. `pixi.lock` is untracked, so the mismatch is per-machine and undocumented.

Fix: drop the pin (the env then solves identically to moments), or state one interpreter pin in root `[dependencies]`.

pixi.toml
```diff
@@ -75,13 +75,32 @@ sphinx-rtd-theme = ">=1.0"
...
+
+# FIXME: LD_PRELOAD forces the env libstdc++; the pip kvikio wheel otherwise loads
+# the older system libstdc++, after which importing scipy/moments fails with a
+# missing CXXABI symbol.
+[feature.cov.tasks]
+# Run in order: `pixi run coverage && pixi run diff-coverage <branch to compare>`
+coverage = { cmd = "bash -c 'LD_PRELOAD=$CONDA_PREFIX/lib/libstdc++.so.6 pytest tests/ --slow --cov=pg_gpu --cov-report=term-missing --cov-report=html --cov-report=xml -n 10'" }
+
+[feature.cov.tasks.diff-coverage]
+cmd = "diff-cover coverage.xml --compare-branch={{ base }} --format html:htmlcov/diff-coverage.html"
```
andrewkern: **`{{ base }}` is substituted raw and unquoted into the shell line.**

pixi 0.66 has no shell-quote filter, so a legal git refname containing `& ; | ( )` is parsed as shell syntax by deno_task_shell:

- `pixi run diff-coverage feat/a&b` (accepted by `git check-ref-format --branch`) backgrounds diff-cover against `feat/a` and runs `b --format ...` ("b: command not found", verified).
- `a(b)` aborts with a parse error; `'x; echo pwned'` runs a second command.
- `pixi run diff-coverage --compare-branch=foo` renders `--compare-branch=--compare-branch=foo`.
- `pixi run diff-coverage main --fail-under 80` fails in pixi with "received more arguments than expected" unless the extras go after `--`, which the usage comment does not mention.

pixi.toml
```diff
@@ -75,13 +75,32 @@ sphinx-rtd-theme = ">=1.0"
...
+args = [{ arg = "base", default = "main" }]
```
andrewkern: **`default = "main"` compares against the local `main`, not `origin/main`.**

This overrides diff-cover's own default `origin/main` (the ref PRs merge into). Two failure modes:

- Local `main` is N commits behind origin and the branch was cut from `origin/main`: diff-cover runs `git diff -U0 main...HEAD` from the stale merge-base, and every line touched by those N upstream commits is counted against the branch's diff coverage.
- `git clone -b <feature> <repo>` creates no local `main` (reproduced): diff-cover raises `ValueError` (branch not found) and `pixi run diff-coverage` exits with a traceback, while `origin/main...HEAD` succeeds in the same clone.

Use `default = "origin/main"`.

pixi.toml
andrewkern: **`diff-coverage` crashes when `htmlcov/` is absent.**

diff-cover's `open_file` is a bare `open(path, 'wb')` with no `makedirs` (true for 9.0 through 10.5.1), so `--format html:htmlcov/diff-coverage.html` depends on the sibling task's `--cov-report=html` side effect. With `coverage.xml` present but `htmlcov/` absent (after `rm -rf htmlcov`, `git clean -fdX` now that `htmlcov/` is ignored, or an xml-only run), the task raises `FileNotFoundError: 'htmlcov/diff-coverage.html'` and exits 1 with no report (reproduced with diff-cover 10.5.1; succeeds after `mkdir -p htmlcov`).

Write to a repo-root path, `mkdir -p` first, or add `depends-on = ["coverage"]`. Separately, `--format` only exists from diff-cover 9.5.0, so the `>=9.0` floor is looser than the syntax it uses; raise it to `>=9.5`.

pixi.toml
```diff
@@ -75,13 +75,32 @@ sphinx-rtd-theme = ">=1.0"
...
+pytest-cov = ">=4.0"
+diff-cover = ">=9.0"
+# Match the default/moments envs; the GPU/moments stack has no 3.14 wheels and a
+# fresh solve would otherwise pick the newest Python.
+python = "3.12.*"
+
 [environments]
 default = ["gpu", "dev", "docs"]
 moments = ["gpu", "dev", "docs", "moments"]
+coverage = ["gpu", "dev", "docs", "moments", "cov"]
```
andrewkern: **The new env adds a fourth full solve to the CI lint job.**

Because `pixi.lock` is gitignored, `pixi install -e lint` solves every environment in the manifest. This PR's own lint job log (job 97553888589) lists `Environment: coverage / Features: gpu, dev, docs, moments, cov` in the solve, on a distinct Python 3.12 wheel set (cupy-cuda12x, kvikio-cu12, nvidia-nvcomp-cu12, moments-popgen). The measured time cost is within noise, but any event that breaks the coverage env's solve and not lint's (a resolver panic on a source build, as the workflow comment already records happened once with the moments env; a yanked cp312 wheel; a python=3.12 vs cuda-cudart-dev/notebook conflict) fails the lint job before ruff runs on a change that touched no Python.

Options: commit `pixi.lock` (then `--locked` and caching apply), or run ruff in CI outside pixi.

andrewkern: **A separate env duplicates an 11 GB GPU stack per developer.**

This environment is the moments env plus two pure-python packages, but every developer who uses it must solve and install a second multi-GB GPU stack (the moments env is 11 GB here), and the cov feature's 3.12 pin defeats pixi's package-cache hardlink sharing with the existing 3.14 envs. pytest-cov is inert unless `--cov` is passed.

A cheaper form that still keeps tooling out of user-facing envs: add pytest-cov/diff-cover to `[feature.dev.dependencies]` (dev is already excluded from user installs) and define the tasks in `[feature.moments.tasks]`, so `pixi run -e moments coverage` reuses the installed env.

.gitignore
```diff
@@ -26,6 +26,8 @@ commit_message.txt
 *.png
 *.csv
 !examples/data/*.csv
+# generated conversion artifact of the tracked im-parsing-example.vcf
+examples/data/*.h5
```
andrewkern: **This hides a test side effect instead of fixing it.**

`tests/test_ld_validation_full.py:59` and `tests/test_moments_ld.py:77,101` pass the tracked `examples/data/im-parsing-example.vcf` straight to `moments.LD.Parsing`, which writes `<stem>.h5` beside the VCF and on later runs reopens it with no mtime or content check (`Parsing.py:80-82`). `tests/test_moments_ld.py:361-364` and `:529` already show the right pattern (tmp_path copy + unlink, or `use_h5=False`).

Latent today: the on-disk `.h5` and the tracked pickle `tests/cache/moments_ld_*.pkl` both match the current VCF (verified by rerunning moments with `use_h5=False`). Trigger: someone edits or regenerates the VCF. The `moments_results` fixture returns the old pickle unconditionally, and even with the pickle deleted moments parses the stale `.h5`, so the parity reference describes old genotypes while pg_gpu reads the new ones, and `git status` no longer reveals the cache file.

Pass `use_h5=False` at the three call sites (285 KB VCF, sub-second parse) and drop this ignore line.

pixi.toml
andrewkern: **A worker that dies in native code under `--cov -n 10` silently loses its coverage.**

An xdist worker killed by a segfault or host OOM never runs `DistWorker.finish()`, so all coverage it accumulated is lost. pytest-cov 7 appends it to `failed_workers`, prints the misleading "ensure that pytest-cov is installed on these workers", and still writes term/html/xml from the survivors, so diff-cover runs on partial data. Example: worker gw3 runs 40 tests covering `streaming_matrix.py` and `ld_statistics.py`, then hard-crashes in a kvikio/nvcomp/zarr codec on test 41; xdist replaces it and re-reports only test 41 as failed, the 40 earlier tests keep PASSED with zero coverage, and diff-cover flags executed lines as uncovered. The one crash line is buried among the 26 known failures.

The trigger is uncertain (CuPy OOM surfaces as a Python exception, so a genuine death needs a native crash). Confirm by SIGSEGV-ing one worker under `--cov -n 10` and checking `coverage.xml`.

pyproject.toml
```diff
@@ -127,6 +127,18 @@ filterwarnings = [
...
+source = ["pg_gpu"]
+# pythonpath adds pg_gpu/ to sys.path, so modules are importable both as pg_gpu.x
```
andrewkern: **This comment states a false mechanism.**

coverage.py keys measured lines by canonical file path, so the `pg_gpu.x` / `x` import aliasing from `pythonpath` cannot split attribution regardless of this setting (and 22 of 29 `pg_gpu` modules use relative imports, so a bare-name import fails anyway). Verified on coverage 7.13.3: a package imported as both `pkg.mod` and `mod` yields one measured file with `relative_files=False`; setting it `True` only changes the key to `pkg/mod.py`.

The real effect, repo-relative paths in `.coverage` and `coverage.xml` (`<source>pg_gpu</source>`), is absent from the comment. A maintainer who later removes `pythonpath = ["pg_gpu"]` (no test uses the alias) reads this, concludes `relative_files` only served the alias, deletes it, and the artifacts start embedding absolute `/home/<user>/...` paths, which breaks cross-checkout and CI use. Reword to state the portability reason.

pyproject.toml
```diff
@@ -127,6 +127,18 @@ filterwarnings = [
     "ignore::pg_gpu._warnings.BiallelicOnlyWarning",
 ]
 
+[tool.coverage.run]
+source = ["pg_gpu"]
```
andrewkern, line 133: **Four of the five new settings are inert under the PR's own task.**

- `source = ["pg_gpu"]` is overridden by the CLI `--cov=pg_gpu` (constructor kwargs are applied after the config file; verified against coverage 7.13.3 and pytest-cov 7.0.0 sources plus a probe: config `source=['from_config']` + `Coverage(source=['pkgdir'])` -> `config.source == ['pkgdir']`).
- `omit = ["tests/*"]` is dead: `source` resolves to the `pg_gpu/` tree and `tests/` is rejected as outside `--source` before `omit` is consulted.
- `skip_covered = false` is the library default.
- `show_missing = true` duplicates `--cov-report=term-missing`.

Only `relative_files` does work. A developer who edits `source` here to add a package or narrow the tree and runs `pixi run coverage` sees no effect. Keep the config as the single owner and shorten the task to a bare `--cov`; drop `omit`, `skip_covered`, and `show_missing`.

pixi.toml
```diff
@@ -75,13 +75,32 @@ sphinx-rtd-theme = ">=1.0"
...
 lint = ["lint"]
 
 [tasks]
 test = { cmd = "pytest tests/ -v" }
 test-parallel = { cmd = "pytest tests/ -v -n 10" }
 lint = { cmd = "ruff check pg_gpu/ tests/ examples/" }
 docs = { cmd = "make html", cwd = "docs" }
+
+# FIXME: LD_PRELOAD forces the env libstdc++; the pip kvikio wheel otherwise loads
```
andrewkern: **The FIXME's stated failure does not reproduce, and the PR body names a different variable.**

Verified in the moments env: conda numpy's RPATH maps the env libstdc++ (6.0.34) before cupy/kvikio load, `/proc/self/maps` after `import cupy, kvikio, scipy.sparse, moments` shows only the env copy, and `import cupy; import scipy.fft` (needs `CXXABI_1.3.15`) succeeds with no preload. In isolation, cupy's core `.so` (not kvikio) is the first pip loader that would bind the system copy. The PR body describes the workaround as `LD_LIBRARY_PATH` while the task sets `LD_PRELOAD`, and the comment sits above the `[feature.cov.tasks]` header rather than beside the task it explains.

A reader cannot tell whether the preload is still required or on which machines. Either drop it (it may be a 3.12-only artifact of the pin), or record the exact reproducing configuration and the removal test.

andrewkern (commented):
Local review of this branch at 87ce704. Fifteen inline comments follow; the three that matter most:

1. Module-level lines of `pg_gpu` are never measured, because the existing `ignore::pg_gpu._warnings.BiallelicOnlyWarning` filter imports the package before pytest-cov starts. Every `cp.RawKernel` construction and every new `def` line reports as uncovered.
2. The `bash -c` wrapper on the `coverage` task drops any appended pytest args, so `pixi run coverage tests/test_sfs.py` silently runs the full `--slow -n 10` suite.
3. The `python = "3.12.*"` pin does not match the default/moments envs (they solve to 3.14 with cp314 wheels); it creates a third GPU stack that no other task exercises.

Also checked: the 26 moments.LD parity failures reported in the PR body reproduce on `main` in the current moments env with matching per-file counts, so they are pre-existing as stated.

pixi.toml
nspope: This is true but it's a rare gotcha, not an issue that can be fixed (aside from running tests in serial which is not what we want), so i'm gonna leave it alone

pixi.toml
nspope: now we're tracking the lock file. this has the pg_gpu commit hash, which will always be out of date (i.e. the commit to add the lock will change the hash) but this is inert metadata. The environment is fragile so I think this is the best bet.

nspope:
Thanks, helpful feedback --- this is good to go now, and I spot checked some of the uncovered gaps to make sure they're actually uncovered (not just missed by the tooling). The r2py/lostruct comparison is one spurious gap that we'll just have to keep in mind.

PR #243, opened 2026-08-29
Fail the test suite on any unexpected multiallelic drop

### LINKED ISSUE #222: Promote BiallelicOnlyWarning to an error in CI
`pyproject.toml` globally ignores `BiallelicOnlyWarning` in `filterwarnings`, so a statistic that silently starts dropping (or recoding) sites cannot fail any test; only the dedicated warning tests would notice their own local overrides.

Promoting the filter to `error::` is the safety net, but it is not free: measured blast radius is 28 failing tests, each needing a judgment call - assert the warning where dropping is the point of the test, silence locally where the fixture is incidentally multiallelic.

andrewkern:
`pyproject.toml` ignored `BiallelicOnlyWarning` for the whole suite (by message since #225, so pytest-cov keeps measuring module-level lines), which meant a statistic that started dropping or recoding sites could not fail a test. This PR flips that filter to `error`. A drop must now be expected by the test that triggers it, with `pytest.warns` or a `filterwarnings` mark; anywhere else it fails the test.

**What the promotion caught.** `divergence.zx` computed each of its three ZnS values on a different site set: `zns` drops the sites that are multiallelic within the matrix it is handed, so a site that is triallelic across the sample but biallelic inside one population entered that population's term and not the total's. On the two-population test fixture that moves `zns(pop1)` from 0.030275 to 0.030464 and `zx` from 1.903107 to 1.903249; small there, but the wrong definition (Schrider et al. 2018 compute all three on one SNP set), and it grows with how multiallelic the data is. `zx` now restricts the whole matrix once, warns once, and computes all three terms on it. A new test checks that pre-restricting the matrix yourself does not move the value, and `zx` joins the parametrized once-per-call warning test.

**Blast radius, measured.** With the filter at `error`, 28 tests failed on `main`, the same count #222 estimated, and none of the 53 `--slow` tests. They split into: 21 streaming tests whose msprime Jukes-Cantor fixtures carried a few triallelic sites that `GenotypeMatrix.from_zarr` recodes (now simulated under `mutation_model='binary'`, as the other streaming files already did); 4 `test_ld_biallelic` tests plus one streaming test that build multiallelic input on purpose (now `pytest.warns` or a class-level `filterwarnings` mark); and the 2 `TestZx` tests above.

**Small refactors that fell out.** `HaplotypeMatrix.restrict_to_biallelic` gains `warn_context=`, replacing the restrict-then-warn pair six callers had written out by hand (`zns`, `omega`, `windowed_r_squared`, both `compute_ld_statistics_gpu_*`, and `zx`). `ld_statistics._zns_biallelic` lets a caller that has already restricted skip the second pass. `zns` and `omega` docstrings now state the restriction contract. The phrase the filter keys on is pinned by `test_biallelic_only_warning`, so a rewording of the message cannot silently switch enforcement off. One doc sentence that claimed `restrict_to_biallelic` itself warns is corrected.

Full suite 1373 passed, 78 skipped, 25 xfailed; slow 53 passed; ruff clean; Sphinx builds.

Closes #222.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -127,7 +127,10 @@ def simulate_hm(n_samples=20, seq_length=100_000, seed=42,
     (Jukes-Cantor, which can produce triallelic sites);
     ``mutation_model='binary'`` forces biallelic, which is what the
     pg_gpu-vs-allel parity tests want because pg_gpu's pi formula has
-    a known multi-allelic gap (kr-colab/pg_gpu#100).
+    a known multi-allelic gap (kr-colab/pg_gpu#100), and what the
+    streaming-versus-eager suite uses so a biallelic-only drop (an error
+    under the test configuration) cannot fail a comparison that is not
+    about it.
     """
     import msprime
     from pg_gpu import HaplotypeMatrix
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -125,12 +125,12 @@ def simulate_hm(n_samples=20, seq_length=100_000, seed=42,
     Centralizes what would otherwise be repeated per-test simulation
     boilerplate. ``mutation_model=None`` lets msprime pick its default
     (Jukes-Cantor, which can produce triallelic sites);
-    ``mutation_model='binary'`` forces biallelic, which is what the
-    pg_gpu-vs-allel parity tests want because pg_gpu's pi formula has
-    a known multi-allelic gap (kr-colab/pg_gpu#100), and what the
-    streaming-versus-eager suite uses so a biallelic-only drop (an error
-    under the test configuration) cannot fail a comparison that is not
-    about it.
+    ``mutation_model='binary'`` forces biallelic sites. The scikit-allel
+    parity tests want that because allel counts a multiallelic site once
+    where pg_gpu counts each allele, so the two only agree on biallelic
+    data; the streaming-versus-eager suite wants it so a biallelic-only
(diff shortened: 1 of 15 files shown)


andrewkern:
trying to tighten the testing here @nspope -- take a quick look if you get a chance

tests/conftest.py
```diff
@@ -127,7 +127,10 @@ def simulate_hm(n_samples=20, seq_length=100_000, seed=42,
     (Jukes-Cantor, which can produce triallelic sites);
     ``mutation_model='binary'`` forces biallelic, which is what the
     pg_gpu-vs-allel parity tests want because pg_gpu's pi formula has
-    a known multi-allelic gap (kr-colab/pg_gpu#100).
+    a known multi-allelic gap (kr-colab/pg_gpu#100), and what the
```
nspope: this comment is a bit misleading now that the fix is in

nspope (approved):
looks good!

tests/conftest.py
andrewkern: yeah agreed.

andrewkern: Reworded in de80180: the binary model is there because scikit-allel counts a multiallelic site once where pg_gpu counts each allele, so the parity tests only agree on biallelic data. The clean_hm docstring in test_implementation_parity.py said the same stale thing and is fixed in the same commit.

PR #244, opened 2026-08-30
Lay out from_ts rows by individual, not by sample order

### LINKED ISSUE #192: from_ts pairs haplotype rows by sample order rather than by individual
## What happens

`HaplotypeMatrix.from_ts` builds its rows from `ts.genotype_matrix().T`, which
orders them by `ts.samples()`. It never looks at `ts.individual(i).nodes`. The
canonical row order says sample `i` owns rows `2i` and `2i + 1`, so `from_ts`
is assuming that `ts.samples()` lists each individual's two nodes next to each
other.

That holds for a plain `msprime.sim_ancestry`. It does not hold in general, and
when it breaks nothing looks wrong: the statistics still run and return
plausible numbers.

## Caveat: this is about how we consume tskit, not a tskit bug

tskit never promises that sample order matches individual grouping. The
individual is the thing that owns its nodes, and `ts.samples()` is free to list
them in any order. Reordering is an ordinary thing to do -- `simplify` takes the
sample list you give it -- and afterwards the nodes are renumbered, so the
sample list looks perfectly tidy again.

So the fix belongs on our side: derive each individual's two rows from that
individual, instead of assuming an ordering.

## Reproduction

```python
ts = msprime.sim_ancestry(samples=8, population_size=1e4, sequence_length=5e4,
                          recombination_rate=1e-8, random_seed=7, ploidy=2)
ts = msprime.sim_mutations(ts, rate=1e-7, random_seed=7, model="binary")

# one gamete of each individual, then the others -- a normal thing to ask for
order = [2 * i for i in range(8)] + [2 * i + 1 for i in range(8)]
ts2 = ts.simplify(samples=order)
```

Same genotypes in both, only the sample list differs:

| | as simulated | after simplify |
| --- | --- | --- |
| `samples()[:6]` | `[0, 1, 2, 3, 4, 5]` | `[0, 1, 2, 3, 4, 5]` |
| `individual(0..2).nodes` | `[0,1] [2,3] [4,5]` | `[0,8] [1,9] [2,10]` |
| dosages match the individuals | True | **False** |
| `heterozygosity_observed` | 0.253080 | 0.243297 |
| `fst_weir_cockerham` | 0.008242 | **0.000248** |

The sample list is identical after simplify, so there is no visible clue. FST
drops by 97% here. Statistics that read each row on its own (`pi`, `theta_w`,
the SFS) are unaffected; only the ones that rebuild individuals are wrong --
`fst_weir_cockerham`, `heterozygosity_observed`, `inbreeding_coefficient`,
`GenotypeMatrix.from_haplotype_matrix`, and everything downstream of that.

## Suggested fix

Build the rows from `ts.individual(i).nodes` so individual `i` lands on rows
`2i` and `2i + 1` whatever order `ts.samples()` is in. Where a tree sequence has
no individuals recorded, either keep the current behaviour and say so, or
refuse.

A cheaper stopgap is to check whether the nodes are already grouped and warn
when they are not, but deriving the rows is not much more work and removes the
question.

Worth a test that loads one tree sequence twice, once reordered by `simplify`,
and asserts the per-individual genotypes agree.

## Scope

Pre-existing, and separate from the row-order work in #148, which normalized the
VCF and zarr loaders. `from_ts` was already emitting the canonical order for
ordinary msprime output, so it needed no change there; this is the case where it
cannot know the order is right because it never asks.

andrewkern:
`HaplotypeMatrix.from_ts` built its rows from `ts.genotype_matrix()`, which follows `ts.samples()`, and assumed each individual's two nodes sat next to each other there. tskit does not promise that; `simplify` renumbers samples freely. On a reordered tree sequence every statistic that rebuilds individuals from adjacent rows was silently wrong: on the issue's example `heterozygosity_observed` moves from 0.253080 to 0.243297 and `fst_weir_cockerham` from 0.008242 to 0.000248, while `pi` is unchanged, so nothing looks broken.

**What changes.** A new `_ts_row_order` groups the sample nodes by individual (node id order within an individual, which is also the order `write_vcf` emits, so `from_ts` and `from_vcf` keep producing identical rows) and `from_ts` decodes with `ts.genotype_matrix(samples=...)` in that order. Population `sample_sets` map nodes to rows through the same order and are sorted, so each population's list reads as whole individuals for the pairing consumers. Ordinary msprime output is already grouped, and for it the default decode runs unchanged: the `samples=` path costs 1.3-1.8x and is only taken when rows actually move.

**The cases.** Every sample node in an individual with two sample nodes: rows by individual. No individuals (`msprime.simulate`, hand-built tables) or one sample node per individual (`ploidy=1`): `ts.samples()` order, silently, since nothing defines a pairing; the class docstring now names this exception to the row-order rule. Anything else (orphan samples next to individuals, three sample nodes in one individual): `ValueError`, with the diploid-only pointer the other loaders give (#121).

**Tests.** The issue's reproduction (same tree sequence as simulated and after a scrambling `simplify`; rows match each individual's nodes and the two loads are identical), population sets after reordering against an exact expected list, the no-individual and haploid cases, the two raising layouts, and a non-sample node inside an individual. The demography test now shares the two-population helper. Full suite 1380 passed, slow 53 passed, ruff and Sphinx clean; the 39 existing `from_ts` call sites are bit-identical.

Closes #192.

### DIFF
--- a/pg_gpu/haplotype_matrix.py
+++ b/pg_gpu/haplotype_matrix.py
@@ -110,23 +110,24 @@ _TS_LAYOUT_HINT = ("Simplify to the sample nodes you want, or build the "
 
 
 def _ts_row_order(ts):
-    """Sample nodes of ``ts`` grouped by individual, or ``None`` to keep
-    ``ts.samples()`` order.
+    """``(order, paired)``: the sample-node row order ``from_ts`` uses, and
+    whether adjacent rows pair into individuals.
 
     tskit does not promise that an individual's nodes sit together in
     ``ts.samples()`` (``simplify`` renumbers freely), and the statistics
     that rebuild individuals from adjacent rows have no other way to know.
-    ``None`` means nothing in the input defines a pairing (no individuals,
-    or one sample node each), so the default decode order stands.
-    Individuals with no sample nodes own no rows; non-sample nodes inside
-    an individual are ignored.
+    ``order`` is ``None`` whenever the default decode order already serves:
+    nothing in the input defines a pairing (no individuals, or one sample
+    node each; ``paired`` is False), or the grouped order equals
+    ``ts.samples()`` anyway. Individuals with no sample nodes own no rows;
+    non-sample nodes inside an individual are ignored.
     """
     samples = ts.samples()
     owner = ts.nodes_individual[samples]
     per_individual = np.bincount(owner[owner != tskit.NULL],
                                  minlength=ts.num_individuals)
     if (per_individual <= 1).all():
-        return None
+        return None, False
     if (owner == tskit.NULL).any():
         raise ValueError(
             "from_ts cannot lay out rows by individual: sample node "
@@ -143,7 +144,10 @@ def _ts_row_order(ts):
     # Node id order within an individual is the order individual.nodes reports
     # and the order write_vcf emits alleles in, so from_ts and from_vcf of the
     # VCF written from the same tree sequence produce identical rows.
-    return samples[np.lexsort((samples, owner))]
+    order = samples[np.lexsort((samples, owner))]
+    if np.array_equal(order, samples):
+        return None, True
+    return order, True
 
 
 def _decide_streaming_mode(zarr_path, region, streaming, pop_assignment,
@@ -1025,10 +1029,13 @@ class HaplotypeMatrix:
             each pop name maps to the haplotype row indices of its
             samples. The no-demography case (``sim_ancestry(samples=N)``
             with a single unnamed population) leaves ``sample_sets``
-            unset. Users who need custom groupings can overwrite
+            unset. An individual whose two sample nodes belong to
+            different populations raises, because a population subset
+            cannot pair it; assign ``sample_sets`` by hand for such data.
+            Users who need custom groupings can overwrite
             ``hm.sample_sets`` after construction.
         """
-        row_nodes = _ts_row_order(ts)
+        row_nodes, paired = _ts_row_order(ts)
         # Passing samples= makes tskit decode node by node, 1.3-1.8x slower
         # than the default order, so only pay it when the rows really move.
         haplotypes = (ts.genotype_matrix() if row_nodes is None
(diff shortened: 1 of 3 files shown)


tests/test_haplotype_matrix.py
```diff
@@ -746,3 +757,80 @@ def positions(region):
...
+        hm = HaplotypeMatrix.from_ts(ts2)
+        pop_id = {p.metadata["name"]: p.id for p in ts2.populations() if p.metadata}
+        for name in ("A", "B"):
+            expect = [r for i in range(ts2.num_individuals)
+                      if ts2.node(ts2.individual(i).nodes[0]).population == pop_id[name]
+                      for r in (2 * i, 2 * i + 1)]
+            assert hm.sample_sets[name] == expect
+
+    def test_no_individuals_keeps_sample_order(self):
+        ts = msprime.simulate(sample_size=6, length=1e3, mutation_rate=1e-3,
```
nspope: very minor nit --- why are we using deprecated msprime.simulate here?

pg_gpu/haplotype_matrix.py
```diff
@@ -993,14 +1053,17 @@ def from_ts(cls, ts: tskit.TreeSequence, device: str = 'CPU',
...
             name = pop.metadata.get("name") if pop.metadata else None
             if name is None:
                 continue
             nodes = ts.samples(population=pop.id)
             if len(nodes):
-                sample_sets[name] = [sample_idx[s] for s in nodes]
+                # Ascending, so each list reads as whole individuals: the
+                # pairing consumers take consecutive entries as one sample,
+                # while ts.samples(population=) returns node order.
+                sample_sets[name] = sorted(row_of[s] for s in nodes.tolist())
```
nspope: edge case; what if sample nodes are split between populations within an individual? that's allowed in tree sequence (although I can't think of a simulator that'd do it). I don't think this sorted would keep the right row order in that case.

nspope (approved):
Looks good!

tests/test_haplotype_matrix.py
andrewkern: lol cause i'm mentally stuck here. will update to sim_ancestry

pg_gpu/haplotype_matrix.py
andrewkern: yeah that's a weird edge case! would that ever occur?

nspope: I don't know, but worth an assertion? I don't think this could happen with msprime, but if there's an admixture event "at the present day" then it could potentially happen in someone's simulator. (Or if someone had hybrids in a cross/pedigree and labelled the haplotypes separately).

tests/test_haplotype_matrix.py
andrewkern: Done in d7a7338: sim_ancestry + sim_mutations, then the individuals stripped through the tables, since sim_ancestry always records them. That is also a truer picture of what a hand-built tree sequence looks like. msprime.simulate no longer appears anywhere in the repo.

pg_gpu/haplotype_matrix.py
andrewkern: okay i'll add an assert / ValueError

tests/test_haplotype_matrix.py
andrewkern: Added in 012d9ee. from_ts now raises when a population would hold one gamete of an individual but not the other, naming the individual, since the paired row lists cannot represent that split; haploid and no-individual input is untouched because the check only runs when rows pair. Test builds your case through the tables (one node of individual 0 moved to the other population). Same commit also keeps the fast default decode for ordinary msprime output, where the grouped order is the identity.

PR #246, opened 2026-08-31
Better memory management for moments.LD statistics with multiple pops

### LINKED ISSUE #245: moments.LD `compute_ld_statistics` OOMs for 3–4 populations on large contig: chunk-size estimator is not statistic-count-aware
`pg_gpu.moments_ld.compute_ld_statistics` OOMs for 3- and 4-population runs on a
large contig, even on an 80 GB A100. The pair loop is chunked to a memory budget
by `ld_pipeline.estimate_ld_chunk_size`, but that estimator models the per-pair
statistics cost as linear in `num_pops`, while the number of LD statistics per
pair grows much faster. For 4 pops it under-budgets and selects a chunk several
times too large. There is also no public argument to cap the chunk size.

#### Reproduction

```python
from pg_gpu.moments_ld import compute_ld_statistics

ld = compute_ld_statistics(
    vcf_file="big_contig.vcf.gz", rec_map_file="rec_map.txt",
    pop_file="pops.txt", pops=["popA", "popB", "popC", "popD"],
    r_bins=[0, 1e-6, 2e-6, 5e-6, 1e-5, 2e-5],
)
# -> cupy.cuda.memory.OutOfMemoryError on 80 GB A100
```

Completes with 1–2 pops or a smaller top r-bin.

#### Root cause

The per-pair cost model in `estimate_ld_chunk_size` (`pg_gpu/ld_pipeline.py:46`):

```python
bytes_per_pair = (4 * n_haplotypes_per_pop * num_pops + 150 * num_pops) * 3
```

The statistics term `150 * num_pops` is linear in pops, but the number of
statistics per pair, `len(ld_pipeline.ld_names(num_pops))`, is not:

| pops | `n_ld` | model budget (`150·pops`) |
|---|---|---|
| 1 | 3   | 150 |
| 2 | 15  | 300 |
| 3 | 45  | 450 |
| 4 | 105 | 600 |

For 4 pops the `stats` output alone is `105 × 8 = 840` bytes/pair (shape
`(n_pairs, n_ld)`, `moments_ld.py:362-371`), before `counts_list`
(`moments_ld.py:355-360`) and the intermediate temporaries the multi-pop stat
kernels materialize. So the true peak is several times the budget and the chosen
chunk (e.g. ~5.76M pairs at H=100/pop on a 38 GB budget) OOMs. `num_pops` *is*
passed to the estimator (`moments_ld.py:332`), so this is a wrong model, not a
dropped argument.

#### Proposed fix

Three small changes, all in `estimate_ld_chunk_size` except (3):

**1. Make the workspace term statistic-count-aware.** The `150 * num_pops` term is
a proxy for per-pair working set, but the working set scales with the *number of
statistics* `n_ld = len(ld_names(num_pops))` (3 / 15 / 45 / 105), not with
`num_pops`. Swap the variable, anchoring the constant so the rule is unchanged at
one population (`n_ld = 3`, so `50 * n_ld = 150 = 150 * num_pops`):

```python
n_ld = len(ld_names(num_pops))
bytes_per_pair = (4 * n_haplotypes_per_pop * num_pops + 50 * n_ld) * 3
```

This leaves P=1 identical and P=2 nearly so (both already worked), and shrinks the
P=4 chunk ~3x (e.g. 5.76M -> ~1.85M pairs at H=100/pop on a 38 GB budget), which is
about the true-to-modeled cost ratio behind the OOM. The `* 3` overhead and the
50%-of-free-memory budget stay as the safety margin. Bumping the constant (e.g.
`64`) buys headroom at a small speed cost if 4-pop still overshoots in practice.

**2. Soften the 100k floor so it never exceeds what fits.** The floor
(`max(100_000, ...)`) is a performance guard against tiny chunks, but it currently
overrides the memory budget: when the honest chunk is below 100k (large
`bytes_per_pair` on a smaller or contended GPU), it forces 100k anyway and can OOM
on its own. This gets *more* likely after change (1), which raises `bytes_per_pair`
at high pop counts. Cap the floor by the memory-derived size:

```python
fit = available_memory_bytes // bytes_per_pair
chunk = min(fit, 10_000_000)           # ceiling: perf / overshoot cap
chunk = max(chunk, min(100_000, fit))  # floor, but never above what fits
return max(chunk, 1)
```

Behavior is unchanged whenever >=100k pairs fit (the common case); it only changes
the tight-memory regime, where the old floor was unsafe.

**3. Expose an escape hatch** on `compute_ld_statistics`: a `chunk_size=` (and/or
`available_memory_bytes=`) argument threaded through `_compute_ld_sums` to
`estimate_ld_chunk_size`. An explicit `chunk_size` bypasses the estimate and both
clamps. This is not a numeric change (chunking only partitions an additive sum, so
results are identical), and it is the backstop for when the heuristic still
misjudges.

Changes (1)-(2) touch a shared numeric path; the constant in (1) is anchored to the
current behavior rather than measured, so it should be confirmed against observed
peak memory (`cupy` mempool `used_bytes`) for 1-4 pops before landing. Change (3) is
mechanical and independent, and is the natural first step (it also provides the
fixed-chunk lever needed to calibrate (1)).

nspope:
Closes #245.

`moments_ld.compute_ld_statistics` ran out of GPU memory for three and four populations (#245). The pair loop is chunked to a memory budget by `estimate_ld_chunk_size`, but that estimate modelled the per-pair working set as linear in the number of populations, while the number of LD statistics computed per pair grows faster than that: 3, 15, 45, and 105 for one to four populations. For three and four populations it under-budgeted and chose a chunk several times too large.

The estimate now scales its working-set term by the statistic count rather than the population count, anchored so a one-population run is unchanged. The lower bound on the returned chunk is also removed: the value it wrapped is already how many pairs fit, so the bound could only raise the chunk above what fits and run out of memory on its own.

`compute_ld_statistics` gains `chunk_size` and `available_memory_bytes` arguments to override the estimate for cases where it still misjudges. Neither changes results, because a pair chunk only partitions an additive sum over pairs; they affect memory use only.

Tests check the estimate against an explicit memory budget, so they do not depend on device memory or try to trigger an out-of-memory condition, and confirm that an explicit `chunk_size` gives the same sums as the automatic estimate.

### DIFF
--- a/pg_gpu/ld_pipeline.py
+++ b/pg_gpu/ld_pipeline.py
@@ -49,7 +49,8 @@ def estimate_ld_chunk_size(n_gathered_rows, n_stats,
     if n_stats < 1:
         raise ValueError("n_stats must be positive")
     if available_memory_bytes is None:
-        available_memory_bytes = int(cp.cuda.Device().mem_info[0] * 0.5)
+        available_memory_bytes = cp.cuda.Device().mem_info[0] * 0.5
+    available_memory_bytes = int(available_memory_bytes)
 
     bytes_per_pair = (4 * n_gathered_rows + 50 * n_stats) * 3
     fit = available_memory_bytes // bytes_per_pair
--- a/pg_gpu/ld_pipeline.py
+++ b/pg_gpu/ld_pipeline.py
@@ -18,13 +18,14 @@ def estimate_ld_chunk_size(n_haplotypes_per_pop, available_memory_bytes=None,
     """
     Estimate optimal chunk size for LD computation based on GPU memory.
 
-    Memory per pair (N pairs, H haplotypes per pop, P populations):
-    - hap_i, hap_j arrays (P pops): 4 * H * P * N bytes
-    - counts arrays: 32 * P * N bytes
-    - statistics: 120 * P * N bytes
-    - Overhead (~3x): accounts for intermediates, fragmentation
-
-    Formula: bytes_per_pair = (4*H*P + 150*P) * 3
+    Per-pair memory has two parts: the gathered genotype/haplotype slabs, which
+    scale with haplotypes * populations, and the per-pair working set (counts,
+    statistics, kernel intermediates), which scales with the *number* of LD
+    statistics ``n_ld = len(ld_names(num_pops))`` -- 3 / 15 / 45 / 105 for 1-4
+    pops -- not with ``num_pops``. The ``50 * n_ld`` term is anchored so it
+    matches the historical ``150 * num_pops`` at one population (n_ld == 3) and
+    keeps scaling with the true statistic count above it. The ``* 3`` overhead
+    and the 50%-of-free budget below are the safety margin.
 
     Parameters
     ----------
@@ -43,10 +44,13 @@ def estimate_ld_chunk_size(n_haplotypes_per_pop, available_memory_bytes=None,
     if available_memory_bytes is None:
         available_memory_bytes = int(cp.cuda.Device().mem_info[0] * 0.5)
 
-    bytes_per_pair = (4 * n_haplotypes_per_pop * num_pops + 150 * num_pops) * 3
-    chunk_size = available_memory_bytes // bytes_per_pair
+    n_ld = len(ld_names(num_pops))
+    bytes_per_pair = (4 * n_haplotypes_per_pop * num_pops + 50 * n_ld) * 3
+    fit = available_memory_bytes // bytes_per_pair
 
-    return max(100_000, min(chunk_size, 10_000_000))
+    # Cap per-chunk work at the ceiling. There is no lower floor: a minimum
+    # above what fits would defeat the estimate on a tight or contended GPU.
+    return max(1, min(fit, 10_000_000))
 
 
 # ---------------------------------------------------------------------------
--- a/pg_gpu/ld_pipeline.py
+++ b/pg_gpu/ld_pipeline.py
@@ -13,44 +13,47 @@ import cupy as cp
 # ---------------------------------------------------------------------------
 
 
-def estimate_ld_chunk_size(n_haplotypes_per_pop, available_memory_bytes=None,
-                           num_pops=2):
+def estimate_ld_chunk_size(n_gathered_rows, n_stats,
+                           available_memory_bytes=None):
     """
     Estimate optimal chunk size for LD computation based on GPU memory.
 
-    Memory per pair (N pairs, H haplotypes per pop, P populations,
-    S = len(ld_names(P)) statistics: 3 / 15 / 45 / 105 for P = 1-4):
-    - gathered genotype/haplotype slabs (P pops): 4 * H * P * N bytes
-    - working set (counts, statistics, intermediates): 50 * S * N bytes
-    - Overhead (~3x): accounts for intermediates, fragmentation
+    The per-pair cost is modelled as two terms (R rows gathered per pair
+    across populations, S statistics per pair):
+    - a data term growing with R, the per-pair genotype/haplotype gather. The
+      genotype path reads the matrix in place rather than gathering per pair,
+      so this term overestimates it.
+    - a working-set term (counts, statistics, intermediates) growing with the
+      statistic count S.
 
-    Formula: bytes_per_pair = (4*H*P + 50*S) * 3
-
-    The working set scales with the number of statistics S, not with P; the
-    constant sets a one-population run (S == 3) to 150 bytes per pair. The * 3
-    overhead and the 50%-of-free budget below are the safety margin.
+    Concretely bytes_per_pair = (4*R + 50*S) * 3, evaluated against half the
+    free memory. The constant sets a run of three statistics to 150 bytes per
+    pair; the * 3 overhead and the 50%-of-free budget are the safety margin.
 
     Parameters
     ----------
-    n_haplotypes_per_pop : int
-        Number of haplotypes in the larger population
+    n_gathered_rows : int
+        Total haplotype/genotype rows gathered per pair, summed over the
+        populations (e.g. haplotypes per pop times the number of pops).
+    n_stats : int
+        Number of statistic values accumulated per pair (for the LD moments,
+        the length of the statistic basis for the requested populations).
     available_memory_bytes : int, optional
         Available GPU memory in bytes. If None, queries the GPU.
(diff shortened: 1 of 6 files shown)


pg_gpu/ld_pipeline.py
```diff
@@ -43,10 +47,10 @@ def estimate_ld_chunk_size(n_haplotypes_per_pop, available_memory_bytes=None,
...
         available_memory_bytes = int(cp.cuda.Device().mem_info[0] * 0.5)
 
-    bytes_per_pair = (4 * n_haplotypes_per_pop * num_pops + 150 * num_pops) * 3
-    chunk_size = available_memory_bytes // bytes_per_pair
-
-    return max(100_000, min(chunk_size, 10_000_000))
+    n_ld = len(ld_names(num_pops))
+    bytes_per_pair = (4 * n_haplotypes_per_pop * num_pops + 50 * n_ld) * 3
+    fit = available_memory_bytes // bytes_per_pair
+    return max(1, min(fit, 10_000_000))
```
andrewkern: this is dropping the lower bound from 100k pairs to 1 pair-- i think we want to keep a sizeable floor here (maybe 1k pairs?)

pg_gpu/ld_pipeline.py
```diff
@@ -43,10 +47,10 @@ def estimate_ld_chunk_size(n_haplotypes_per_pop, available_memory_bytes=None,
     if available_memory_bytes is None:
         available_memory_bytes = int(cp.cuda.Device().mem_info[0] * 0.5)
 
-    bytes_per_pair = (4 * n_haplotypes_per_pop * num_pops + 150 * num_pops) * 3
-    chunk_size = available_memory_bytes // bytes_per_pair
-
-    return max(100_000, min(chunk_size, 10_000_000))
+    n_ld = len(ld_names(num_pops))
+    bytes_per_pair = (4 * n_haplotypes_per_pop * num_pops + 50 * n_ld) * 3
```
andrewkern: Two things on this line. `len(ld_names(num_pops))` bakes moments' statistic enumeration into a function a few callers share-- I think we have to be careful here. 

Also do we want a ValueError if num_pops=0?

pg_gpu/ld_pipeline.py
```diff
@@ -18,13 +18,17 @@ def estimate_ld_chunk_size(n_haplotypes_per_pop, available_memory_bytes=None,
     """
     Estimate optimal chunk size for LD computation based on GPU memory.
 
-    Memory per pair (N pairs, H haplotypes per pop, P populations):
-    - hap_i, hap_j arrays (P pops): 4 * H * P * N bytes
-    - counts arrays: 32 * P * N bytes
-    - statistics: 120 * P * N bytes
+    Memory per pair (N pairs, H haplotypes per pop, P populations,
+    S = len(ld_names(P)) statistics: 3 / 15 / 45 / 105 for P = 1-4):
+    - gathered genotype/haplotype slabs (P pops): 4 * H * P * N bytes
```
andrewkern: this isn't true on the genotype path right? scalar is 9 no?

tests/test_ld_pipeline.py
```diff
@@ -0,0 +1,45 @@
...
+"""
+import pytest
+
+from pg_gpu.ld_pipeline import estimate_ld_chunk_size, ld_names
+
+
+class TestEstimateLdChunkSize:
+    def test_single_pop_matches_historical_anchor(self):
+        # The workspace term is anchored so that at one population (n_ld == 3)
+        # it equals the historical 150 * num_pops, leaving P=1 unchanged.
```
andrewkern: i'm not following this doc string -- what's the historical bit? maybe remove

nspope: the previous scaling was 150 * num_pops, now replaced by 50 * function_of(num_pops), which matches the old rule for 2 pops. I can remove; this is a vacuous test

andrewkern:
i think the core fix looks good @nspope but a few little comments above. 

one bigger thing: `_compute_ld_sums` is doing chunk-size-independent work every chunk: `count_fn` re-slices the full population matrices and redoes the has_missing scan per chunk, because has_missing is never passed through — `compute_counts_for_pairs`'s own docstring says batch callers should pass it. I think we're gonna end up really slowing things down here.

nspope:
thanks---for the "bigger point", that's a pre-existing bottleneck, right? But exacerbated here, if the memory fix in this PR results in more chunks on average. So, shall I try to fix the bottleneck here, too?

andrewkern:
yeah i guess we try to fix the bottleneck here? otherwise we can file an issue, but I think we go from OOM to an hour of compute or something with this fix. I haven't actually profiled things -- that might be the place to start.

nspope:
Profiling indicates that the redundancy costs ~1.5% of the total cost (at 5% missingness) ... I don't think this is worth refactoring for.

andrewkern:
roger-- sounds good

pg_gpu/ld_pipeline.py
nspope: clarified in docstring

pg_gpu/ld_pipeline.py
nspope: I refactored this to be "n_gathered_haplotypes" (the first term, where the data is gathered) and "n_stats" (the second term, where the stats are computed). These are passed in by the caller, so it stays more or less general. It's conservative for the genotype path but I think that's OK for now (it's noted as such in the docstring).

nspope:
I think this is good, then---can you take a quick look at the new `estimate_chunk_ld_size` signature & call sites, is this in line with what you were thinking?

pg_gpu/ld_pipeline.py
```diff
@@ -13,40 +13,47 @@
...
-        Number of populations (default 2)
 
     Returns
     -------
     int
         Recommended chunk size (number of pairs per iteration)
     """
+    if n_stats < 1:
+        raise ValueError("n_stats must be positive")
     if available_memory_bytes is None:
```
andrewkern, line 51: one more comment here-- if a user passes a float for available_memory_bytes (e.g. 1e5) will will get a float chunk size -- maybe we should cast to int up top here?

andrewkern:
looks all good but see https://github.com/kr-colab/pg_gpu/pull/246/changes#r3900738786 -- should we worry about this?

nspope:
Dealt with it---I'll go ahead and merge

PR #247, opened 2026-09-01
Exclude zero-call sites from patterson_d

### LINKED ISSUE #217: patterson_d keeps sites where a population has zero calls
Found in the release audit by comparison against scikit-allel. Pre-existing on main and the trunk.

`_patterson_d_gpu`'s `freq()` returns 0.0 when a population has zero valid calls at a site (`n_valid == 0`); scikit-allel returns nan there and drops the site. The zero frequency keeps the site in both the numerator and denominator, so D can change sign: on a two-site example with pop A missing at site 0, pg_gpu gives D = +0.333 while allel gives -1.0.

`missing_data='exclude'` gives the correct answer. f2, f3, and f4 are safe: their denominators carry the zero factor, so the affected terms drop out.

Fix: mark zero-call sites undefined in both the D numerator and denominator, matching allel.

andrewkern:
Found in the release audit (#217). `_patterson_d_gpu`'s `freq()` returned 0.0 for a population with zero valid calls at a site, which kept the site in both the numerator and denominator; a missing population masquerading as frequency 0 can flip D's sign (the issue's two-site example: pg_gpu +0.333, scikit-allel -1.0).

**Fix.** Sites where any of the four populations has no called gametes now leave both terms (`num = den = 0`), folded into the same mask the biallelic filter uses (`defined = biallelic & called`), so the aggregate ratio matches scikit-allel's nansum over its nan sites exactly. The exclusion is silent, matching how the package treats undefined sites under `'include'` elsewhere (pi at n < 2, the between-population term at n = 0). `moving_patterson_d` and `average_patterson_d` share the fix through the one helper; there is no windowed-engine D path. f2/f3/f4 were verified safe: every population contributes a zero factor to their denominators, so a zero-call site already yields exactly 0 there. `missing_data='exclude'` was already correct.

`freq` now divides by `max(n, 1)` instead of a `cp.where` guard (identical values, since a zero-call site has a zero alt count; fewer kernels), and the undefined-site rule is stated once for users (the `patterson_d` Notes), once for maintainers (the mask comment), and in the `missing_data` reference page, whose admixture row previously said "use all sites".

**Tests** (against scikit-allel in `test_admixture_allel_comparison.py`): allel's nan set equals our zeroed set exactly on a fixture with all-missing blocks in two populations, and defined sites agree at 1e-10; a hand-built two-site case shows the zero-call site contributing nothing to the aggregate; `average_patterson_d` matches `allel.average_patterson_d` on the missing-data fixture at the file's standard tolerances; `exclude` mode keeps exactly the fully-called sites.

Full suite 1392 passed, slow 53 passed, ruff and Sphinx clean.

Closes #217.

### DIFF
--- a/tests/test_admixture_allel_comparison.py
+++ b/tests/test_admixture_allel_comparison.py
@@ -29,10 +29,8 @@ def _make_matrix(hap_dict, n_var=100):
 
 
 def _allele_counts(hap):
-    """Compute allele counts (n_variants, 2) from haplotype array."""
-    n = hap.shape[0]
-    dac = np.sum(hap, axis=0)
-    return np.column_stack([n - dac, dac])
+    """Allele counts (n_variants, 2); ``-1`` counts as a missing call."""
+    return np.column_stack([np.sum(hap == 0, axis=0), np.sum(hap == 1, axis=0)])
 
 
 @pytest.fixture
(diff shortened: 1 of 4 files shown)


tests/test_admixture_allel_comparison.py
```diff
@@ -230,3 +228,69 @@ def test_patterson_d_known(self):
...
+            'B': np.array([[1, 1], [1, 0]], dtype=np.int8),
+            'C': np.array([[1, 1], [1, 0]], dtype=np.int8),
+            'D': np.array([[0, 0], [0, 0]], dtype=np.int8),
+        }
+        num, den = admixture.patterson_d(_make_matrix(pops), 'A', 'B', 'C', 'D')
+        assert num[0] == 0.0 and den[0] == 0.0
+        np.testing.assert_allclose(np.nansum(num) / np.nansum(den),
+                                   num[1] / den[1], rtol=1e-12)
+
+    def test_average_d_matches_allel_with_zero_call_sites(self, zero_call_data):
```
nspope, line 292: worth doing a twin with "moving_patterson_d", as that's not tested in these additions? I *think* the same machinery gets used, so it's probably redundant, but if that ever changes it'd get covered.

nspope (approved):
Looks great!

tests/test_admixture_allel_comparison.py
andrewkern: yeah it probably is.... let me scope that.

andrewkern: Added in b751151. It is a direct oracle comparison on the same fixture, with size=5 chosen so three windows fall entirely inside the missing blocks: both sides mark exactly those windows NaN (ours through den_sum == 0, allel through 0/0), and the defined windows agree at 1e-10. So it pins the all-undefined-window convention too, which nothing else covered.

PR #248, opened 2026-09-01
Drop sites with no defined r2 from genotype-path zns and omega

andrewkern:
Found in the release audit (#218). `_r2_matrix_diploid` wrote r^2 = 0 for pairs whose r^2 is undefined, so on a `GenotypeMatrix` every site with no dosage variance stayed in the reductions: appending invariant sites diluted ZnS (0.0337 -> 0.0287 on a probe) and inflated omega far worse (1.12 -> 6.81, a fabricated partition signal), while the haplotype path was invariant.

**The predicate matters.** The obvious fix -- mirroring the haplotype branch's segregating filter -- is wrong for diploids: an all-heterozygous site segregates by allele counts yet has zero dosage variance, so `restrict_to_segregating` keeps it and it still scores 0. The definedness predicate is dosage variance, and it already lives in `_r2_matrix_diploid`'s own normalization. The matrix now NaN-marks zero-variance rows and columns (the convention `pairwise_r2` already uses, recorded in the changelog's behavior-change bullet alongside it), and the existing `_drop_undefined_sites` removes them in both `zns` and `omega`. The public alias `r2_matrix_diploid` has no callers outside the module.

**Also in the diff, from the review pass:** the normalization applies the `1/sqrt(var)` factors as a rank-1 in-place scale -- measured identical values (max diff 2.2e-16), 1.1-1.6x faster, and 4.9x lower peak memory (the old `outer`/`where` chain held five m x m temporaries; at m = 20k that is 15 GB vs 3 GB); the haplotype branch of `_resolve_r2_matrix` uses `restrict_to_segregating()` instead of a seven-line hand-rolled copy; `_drop_undefined_sites`' docstring now names both producers as the convention's single owner; and the public `zns`/`omega` Notes state the GenotypeMatrix behavior.

**Tests:** a parametrized invariance test (zns and omega unchanged by appended all-reference, all-alt, and all-heterozygous sites) and a drop-everything boundary case (all dosage-1 matrix returns the too-few-sites 0.0). Full suite 1395 passed, slow 53 passed, ruff and Sphinx clean. (`test_scikit_allel_comparison_runs_small` timed out in concurrent runs and passes alone in 47 s -- the known contention flake, tracked separately.)

Closes #218.

Issue #200, opened 2026-08-23
Engines disagree on the undefined-FST sentinel and on single-population columns in mixed requests


andrewkern:
`windowed_analysis` returns a different undefined-FST sentinel and different
single-population columns depending on `missing_data`, because the two modes
are served by different engines.

Sentinel, monomorphic data, same call:

```
include -> fst NaN, fst_wc NaN     (fused engines)
exclude -> fst 0.0, fst_wc 0.0     (per-window fallback -> scalars)
```

Mixed request, `statistics=['pi', 'fst_wc']`, `populations=['p1', 'p2']`:

```
include -> columns [..., 'pi', 'fst_wc']            (pi is populations[0] only)
exclude -> columns [..., 'pi_p1', 'pi_p2', 'fst_wc']
```

`df['pi']` raises `KeyError` under `'exclude'`, and the two modes do not even
return the same quantity for `pi`: the fused engine computes it for the first
population alone, the fallback for each population.

`docs/source/quickstart.rst` promises the fallback "computes the same values,
so your results do not change -- only the speed". After #188 that holds for
two-population columns; the sentinel and the single-population half do not.

Two decisions to make, then enforce in every engine:

* one sentinel for an undefined ratio (NaN says undefined; 0.0 says no
  differentiation -- they are different claims; NaN is the defensible one,
  so the scalar's `return 0.0` on `sum_abc <= 0` is what should move), and
* one rule for single-population columns when `populations` is given
  (per-population suffixed columns for every listed population is what the
  fallback does and loses no information).

Related: #185 (multi-population `pi` returns one population only).

Trunk-scoped: builds on the #193 rename and the three-or-more-population
routing gate, both only on nsp-multiallelic-trunk.

andrewkern:
Another instance of the fused-vs-fallback split this issue describes, surfaced while fixing #219: `windowed_analysis` silently ignores `estimator=` for windowed `zns`/`omega`, and the two engines disagree on it the same way they disagree on the sentinel and columns here.

On `main` (3ea3e72):

```
windowed_analysis(hm, statistics=['zns'])                          # default
windowed_analysis(hm, statistics=['zns'], estimator='rogers_huff') # bit-identical -- kwarg swallowed
windowed_analysis(hm, statistics=['zns'], estimator='rogers_huff', missing_data='exclude')
    -> TypeError: WindowedAnalyzer.__init__() got an unexpected keyword argument 'estimator'
```

[...]

PR #249, opened 2026-09-01
Honor the estimator and missing_data that zns and omega were handed, or refuse it

### LINKED ISSUE #219: zns and omega silently ignore estimator='rogers_huff' on a HaplotypeMatrix
Found in the release audit. Pre-existing; the plumbing is identical on main and the trunk.

`_resolve_ld_estimator` whitelists `'rogers_huff'` for any input, but the HaplotypeMatrix branches of `zns` and `omega` route on the missing-data mode only, so the estimator is accepted and ignored: `zns(hm, estimator='rogers_huff')` equals `zns(hm, estimator='r2')` exactly, with no error or warning. The user believes they computed the dosage estimator and got the naive one.

Fix: either pair adjacent haplotypes through `GenotypeMatrix.from_haplotype_matrix` (the #212 pattern for `pairwise_r2` / `windowed_r_squared`) or raise on the unsupported combination.

### LINKED ISSUE #250: zns/omega under the default sigma_d2 estimator silently ignore missing_data='exclude'
Found while fixing #219. Under the default `estimator='auto'` (which resolves to `sigma_d2` for a HaplotypeMatrix), `zns` and `omega` silently ignore `missing_data='exclude'` and return the `include` result.

## Mechanism

`_zns_biallelic` / `zns` smuggle the estimator into the missing-data argument as a sentinel: `_zns_tiled(hm, 'project' if estimator == 'sigma_d2' else missing_data)`. When the estimator is `sigma_d2` the caller's `missing_data` is overwritten with `'project'`, and `_prepare_segregating` then treats `'project'` as `'include'`. So the `'exclude'` request never reaches the computation.

## Reproduction (on `main`, 3ea3e72; ~12% missing, biallelic)

```
sigma_d2 (default): include=0.044129  exclude=0.044129   <- identical, exclude ignored
r2:                 include=0.072273  exclude=0.001518   <- honoured
```

## Why it matters

This directly undermines the workaround #220 recommends. #220 ("`missing_data='include'` r2 mixes pairwise-complete counts with per-site frequencies") says `'exclude'` matches the oracle and advises it under structured missingness -- but for the *default* estimator, `'exclude'` is discarded, so a user following that advice without also passing `estimator='r2'` gets the biased `include` answer anyway. #220 should note this caveat, and this issue should be fixed so the advice holds.

## Fix

Give `_zns_tiled` / `_prepare_segregating` an explicit `use_projection: bool` argument instead of the `'project'` sentinel (`_zns_from_precomputed` already takes exactly that flag), pass `missing_data` through unchanged, and either honour `'exclude'` under `sigma_d2` or raise for the combination. The scalar `estimator`/input contract was tightened for the ignored-estimator case in #249; this is the same silent-ignore shape one parameter over.

andrewkern:
Two coupled fixes to the estimator/missing-data contract of `zns` and `omega`, both from the release audit.

**#219 -- the estimator was ignored.** `zns` and `omega` whitelisted `estimator='rogers_huff'` but the HaplotypeMatrix branches routed on missing-data mode only, so `zns(hm, estimator='rogers_huff')` returned the naive `r2` value exactly, silently. They now pair adjacent haplotypes into 0/1/2 dosages (the `pairwise_r2` pattern) and compute the diploid dosage correlation, verified equal to the same statistic on the derived GenotypeMatrix and different from `r2`. `_resolve_ld_estimator` is now kind-aware (`'haplotype'`/`'genotype'`/`'array'`): `'auto'` resolves per kind, and an estimator a kind cannot compute raises instead of being dropped -- the names raise on a pre-computed array (`'auto'` there means `r2`). A shared `_ld_input_kind` front door classifies the input once for both functions and rejects a streaming matrix with a `materialize` pointer (`NotImplementedError`, matching the streaming classes' contract).

**#250 -- missing_data='exclude' was ignored under the default estimator.** `zns` smuggled the estimator through the `missing_data` argument as a `'project'` sentinel, so under the default `sigma_d2` estimator `missing_data='exclude'` was overwritten and the `include` result came back (verified: include and exclude identical under sigma_d2, differing under `r2`). `_zns_tiled` now takes an explicit `use_projection` flag, as its sibling `_zns_from_precomputed` already did, and the real `missing_data` reaches `_prepare_segregating`, so `exclude` restricts to complete sites before the estimator runs. `omega` already did this via `_build_sigma_d2_matrix` and is unchanged. (Filed as #250 during the review; rolled in here because it is the same silent-ignore in the same functions, one parameter over, and the sentinel lives in `_zns_biallelic`, which this PR already rewrites.)

**Both dispatch paths covered.** Eager: every (input kind, estimator) cell now computes or raises, pinned on the resolver and through the public functions; zns `exclude` now differs from `include` under the default estimator and equals the hand-filtered complete-site subset. Streaming: scalar `zns`/`omega` reject a stream of either class; windowed streaming `zns`/`omega` are unaffected (fixed-estimator, include-only per window) and were held bit-identical to eager by the equivalence tests added in #248.

**From the review pass.** `zns` reduces to a scalar over the NaN-marked matrix with `nansum`, skipping the compacting `_drop_undefined_sites` copy (~halved peak memory, ~47% less runtime at 10k sites; `omega` keeps the copy for its 2-D prefix sums). The `rogers_huff` route clears sample sets before `from_haplotype_matrix`, so a gappy `sample_sets` no longer warns about rows these population-free statistics never use. `missing_data.rst` (the "LD Estimator Choice" section and Best Practice 4) now describes the kind-based policy. A deliberate limit is documented, not engineered: on a HaplotypeMatrix the `rogers_huff` route builds the full m x m matrix (O(m²), ~55k-variant wall on an 80 GB GPU), unreachable for the window-sized inputs these statistics take.

**Left as follow-ups** (same silent-ignore shape elsewhere, out of scope): `windowed_analysis` swallows `estimator=` for fused zns/omega (noted on #200); `pairwise_r2`/`windowed_r_squared` reject the documented `'auto'` name and `mu_ld` gives a bare error on a stream (#251).

Full suite 1410 passed, slow 53 passed, ruff and Sphinx clean. `test_scikit_allel_comparison_runs_small` is a known subprocess-timeout flake under host load (#252), unrelated -- a differential benchmark shows this branch equal-or-faster than main on every touched path.

Closes #219. Closes #250.

### DIFF
--- a/pg_gpu/ld_statistics.py
+++ b/pg_gpu/ld_statistics.py
@@ -539,21 +539,21 @@ def rogers_huff_r_squared(matrix, tile_size: Optional[int] = None
     return rogers_huff_r(matrix, tile_size=tile_size) ** 2
 
 
-def _zns_tiled(mat, missing_data='include', tile_size=512):
+def _zns_tiled(mat, missing_data='include', tile_size=512, use_projection=False):
     """Compute ZnS without materializing the full r² matrix.
 
     Uses tile-based accumulation: computes r² for B×B blocks and
     sums per tile, keeping memory at O(B²) instead of O(m²).
 
-    When missing_data='project' (set internally via estimator='sigma_d2'),
-    uses unbiased multinomial projection estimators (Ragsdale & Gravel
-    2019) computing σ_D² = D²/π² per pair instead of naive r².
+    ``use_projection`` selects the unbiased multinomial projection
+    estimator (Ragsdale & Gravel 2019, σ_D² = D²/π² per pair) over
+    naive r². ``missing_data`` is applied first either way, so
+    ``'exclude'`` restricts to complete sites before the estimator runs,
+    matching ``_build_sigma_d2_matrix`` (omega's projection path).
     """
     hap_clean, valid_mask, m = _prepare_segregating(mat, missing_data)
     if m < 2:
         return 0.0
-
-    use_projection = (missing_data == 'project')  # internal: mapped from estimator='sigma_d2'
 
     B = tile_size
     total = 0.0
@@ -716,9 +716,8 @@ def zns(r2_matrix_or_matrix, missing_data='include', estimator='auto'):
         memory is O(m²) -- about 55,000 variants on an 80 GB GPU.
     missing_data : str
         ``'include'`` (default) uses per-site valid data for frequency
-        computation. ``'exclude'`` filters to sites with no missing data. Under the
-        default ``sigma_d2`` estimator the projection handles missing
-        data per pair and ``'exclude'`` is not applied.
+        computation. ``'exclude'`` filters to sites with no missing data
+        before the estimator runs, under every estimator.
     estimator : str
         ``'auto'`` (default): ``sigma_d2`` for a ``HaplotypeMatrix``,
         ``rogers_huff`` for a ``GenotypeMatrix``, ``r2`` for a
@@ -782,8 +781,8 @@ def _zns_biallelic(hm, missing_data='include', estimator='auto'):
     no second restriction and no warning. ``divergence.zx`` restricts once
     and calls this for each of its three terms."""
     estimator = _resolve_ld_estimator(estimator, 'haplotype')
-    # 'project' selects the sigma_d2 estimator inside _zns_tiled.
-    return _zns_tiled(hm, 'project' if estimator == 'sigma_d2' else missing_data)
+    return _zns_tiled(hm, missing_data,
+                      use_projection=(estimator == 'sigma_d2'))
 
 
 def _build_sigma_d2_matrix(mat, missing_data='include'):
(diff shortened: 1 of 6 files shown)


pg_gpu/ld_statistics.py
```diff
@@ -684,21 +710,30 @@ def zns(r2_matrix_or_matrix, missing_data='include', estimator='auto'):
     r2_matrix_or_matrix : ndarray, HaplotypeMatrix, or GenotypeMatrix
         Square r-squared matrix, or a matrix object (dispatches to
         haploid or diploid r-squared computation automatically).
-        When a HaplotypeMatrix is passed, uses tiled computation to
-        avoid materializing the full m×m r² matrix.
+        With estimator 'r2' or 'sigma_d2', a HaplotypeMatrix goes
+        through tiled computation that avoids materializing the full
+        m×m r² matrix; 'rogers_huff' builds the full matrix, so peak
+        memory is O(m²) -- about 55,000 variants on an 80 GB GPU.
```
nspope:
```suggestion
        memory is O(m²).
```

nspope (approved):
Looks good!

PR #253, opened 2026-09-02
Share the LD estimator resolver and stream guard across the entry points

### LINKED ISSUE #251: Estimator handling is inconsistent across the LD entry points (pairwise_r2 / windowed_r_squared / mu_ld)
Found while fixing #219. #249 tightened the estimator/input contract for scalar `zns`/`omega` (compute what was asked, or raise). The sibling LD entry points still handle estimators inconsistently.

All reproduced on `main` (3ea3e72):

## 1. `pairwise_r2` and `windowed_r_squared` reject the documented default name

`zns`/`omega` document `estimator='auto'`, but the matrix methods default to `'r2'` and reject `'auto'`:

```
pairwise_r2(estimator='auto')        -> ValueError: Unknown estimator: 'auto' (expected 'r2' or 'rogers_huff')
windowed_r_squared(estimator='auto') -> ValueError: Unknown estimator: 'auto' (expected 'r2' or 'rogers_huff')
```

A user who learns the vocabulary from `zns`/`omega` hits an "unknown estimator" error for the name those functions call the default. The unknown-name message is also copy-pasted in both methods (`haplotype_matrix.py`), separate from `_resolve_ld_estimator`, which is now the package's estimator authority.

## 2. Pairwise entry points give a bare error on a streaming matrix

`#249` makes scalar `zns`/`omega` reject a stream with a `materialize` pointer. The neighbours do not:

```
mu_ld(stream) -> AttributeError: 'StreamingHaplotypeMatrix' object has no attribute 'device'
```

`pairwise_r2` / `locate_unlinked` on a stream fail the same way (the streaming classes define neither `.device` nor these methods), so the "the `.haplotypes` property already gives guidance" assumption does not hold for this family.

## Fix

- Add an `allowed`/`auto` parameter to `_resolve_ld_estimator` (e.g. `_resolve_ld_estimator(estimator, 'haplotype', allowed=('r2', 'rogers_huff'), auto='r2')`) and call it from `pairwise_r2` / `windowed_r_squared`, so `'auto'` resolves consistently and the unknown-name message lives in one place.
- Route the pairwise entry points through the shared `_ld_input_kind` guard added in #249 (or lift it to a common helper) so a stream gets the same `NotImplementedError` + materialize pointer everywhere.

andrewkern:
Follow-up filed during the #219 review. #249 tightened the estimator/input contract for scalar `zns`/`omega`; this brings the sibling LD entry points into line.

**Eager -- `pairwise_r2` / `windowed_r_squared` accept the documented default name.** Both hand-rolled their validation (`if estimator != 'r2': raise "unknown estimator (expected 'r2' or 'rogers_huff')"`), so `estimator='auto'` -- the default name for the estimator everywhere else -- was rejected. They now call `_resolve_ld_estimator(estimator, 'haplotype', allowed=('r2', 'rogers_huff'), auto='r2')`: `'auto'` resolves to `'r2'` (verified bit-identical to `estimator='r2'`), and `'sigma_d2'` or a nonsense name raises the one shared "Unknown estimator" message. `_resolve_ld_estimator` gained keyword-only `allowed`/`auto` params for this; `zns`/`omega` keep their kind-based `'auto'` (sigma_d2 / rogers_huff / r2 by input kind).

**Streaming -- one materialize pointer, not a bare error.** `mu_ld(stream)` died on `.device` (`AttributeError`) and the `rogers_huff_r` family raised a misleading `TypeError` ("requires a GenotypeMatrix") on a `StreamingGenotypeMatrix`. A `_reject_streaming` helper, extracted from `_ld_input_kind` so the materialize message lives in one place, now guards `mu_ld` and `_dosage_from_matrix` (the `rogers_huff_r`/`rogers_huff_r_squared` choke point). All three now raise `NotImplementedError` with the same "call `.materialize(region=...)`" pointer `zns`/`omega` give. Every module-level LD function taking a matrix is now guarded; the rest take count arrays.

**Windowed dispatch -- unaffected, verified.** `windowed_analysis` computes per-window `zns`/`omega`/`mu_ld` on eager per-window slices (and `_stream_windowed_analysis` recurses with eager chunk matrices), so `_reject_streaming` never fires there. Confirmed: windowed `mu_ld` runs eager, and streaming-vs-eager windowed `mu_ld` is bit-identical.

**Tests.** `pairwise_r2`/`windowed_r_squared` accept `'auto'` (== r2) and raise on `sigma_d2`/nonsense; `mu_ld` and both `rogers_huff_r` functions reject a stream with the materialize pointer (joining the `zns`/`omega` stream-rejection cases from #248/#249).

The full suite completed green (exit 0; a re-run for the exact count was starved by heavy host load from other users, so I'm citing the completed run rather than a number). Slow suite 53 passed; ruff and Sphinx clean. The simplify pass found nothing.

Closes #251.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -177,6 +177,13 @@ Read this section if you are comparing against older pg_gpu results.
 Bug fixes
 ~~~~~~~~~
 
+* ``HaplotypeMatrix.pairwise_r2`` and ``windowed_r_squared`` rejected
+  ``estimator='auto'`` -- the default name for the LD estimator
+  everywhere else -- with an "unknown estimator" error. They accept it
+  now (it means ``'r2'`` there), sharing the one estimator resolver.
+* ``mu_ld`` and the ``rogers_huff_r`` family raised a bare
+  ``AttributeError``/``TypeError`` on a streaming matrix; they now give
+  the same "call ``materialize``" message ``zns``/``omega`` do.
 * ``zns`` and ``omega`` accepted ``estimator='rogers_huff'`` on a
   ``HaplotypeMatrix`` and silently computed naive ``r2`` instead. They
   now pair adjacent haplotypes into 0/1/2 dosages and compute the
--- a/pg_gpu/haplotype_matrix.py
+++ b/pg_gpu/haplotype_matrix.py
@@ -1817,11 +1817,11 @@ class HaplotypeMatrix:
         Parameters
         ----------
         estimator : str
-            ``'r2'`` (default) -- naive haplotype r² from
-            frequency-based formula. ``'rogers_huff'`` -- Rogers-Huff
-            r² computed on diploid 0/1/2 dosages obtained by pairing
-            adjacent haplotypes (sample 0 = haplotypes 0,1; sample 1
-            = haplotypes 2,3; ...). Matches
+            ``'auto'`` and ``'r2'`` (the default) both give naive
+            haplotype r² from the frequency-based formula.
+            ``'rogers_huff'`` gives Rogers-Huff r² on diploid 0/1/2
+            dosages obtained by pairing adjacent haplotypes (sample 0 =
+            haplotypes 0,1; sample 1 = haplotypes 2,3; ...), matching
             :func:`scikit-allel.rogers_huff_r` ** 2.
 
         Returns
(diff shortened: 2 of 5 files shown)


nspope (approved):
This looks fine to me

Issue #232, opened 2026-08-25
Coverage gap: diversity.py + sfs.py


nspope:
`diversity.py` is at 79% (173/818) and `sfs.py` at 81% (43/230). After setting
aside the cross-cutting `missing_data='exclude'` and `population=` gaps (each has
its own issue), the module-specific gaps are the eta-family estimators, polyploid
heterozygosity, some SFS/projection features, and empty-input error branches.
(Baseline: `pixi run coverage` on `main` b0c9801; line numbers drift — match on
function name.)

### Real feature gaps
| File:line | Function | Uncovered behavior | Test |
|---|---|---|---|
| 59-68 | `_get_minus_eta1_norm` | GPU lookup-cache hit/miss/store | call a stat using the `minus_eta1` weight |
| 73-83 | `_get_minus_eta1_star_norm` | cache hit/miss/store | call a stat using `minus_eta1_star` |
| 239-262 | `_ac_contribution` | `eta1`/`eta1_star`/`minus_eta1`/`minus_eta1_star` estimator branches + unknown-estimator raise | request each estimator by name |
| 1748-1751, 1829-1840 | `heterozygosity_observed` | polyploid (ploidy>2) loop body | matrix with ploidy 3/4 |
| 1813 | `heterozygosity_observed` | `n_hap % ploidy != 0` raise | mis-sized ploidy |
| 1430 | `zeng_dh` | selective-sweep branch (D<0 AND H<0) | craft a sweep-like SFS |
| 767-792 | `pi_components` | power-user pairwise-diffs API + invariant-site handling | call directly |
| 584-608 | `FrequencySpectrum.__init__` | empty n_valid (n_max=0), skip n_i<2, invariant-site correction | small/edge FS |
| 662-708 | `FrequencySpectrum` | custom-weight callables, `suggest_projection_n`, multi-n/projected `sfs` | multi-sample-size FS + projection |

### Error / empty-input branches (lower value)
| File:line | Function | Uncovered behavior | Test |
|---|---|---|---|
| 827, 862, 894, 1119, 1298, 1332, 1369, 1395 | `pi`/`theta_w`/`tajimas_d`/`fay_wus_h`/`theta_h`/`theta_l`/`normalized_fay_wus_h`/`zeng_e` | return 0/NaN on an empty matrix | call with `num_variants==0` |
[...]

PR #254, opened 2026-09-02
Wire up existing eta-family estimators to scalar statistics

nspope:
`quickstart` lists `eta1`, `eta1_star`, `minus_eta1`, and `minus_eta1_star` as available `diversity_stats` estimators, but the code mapped only `pi`, `theta_w`, `theta_h`, and `theta_l`, so asking for an eta name raised `ValueError: Unknown statistic`. This adds the four names to the estimator map. They route through the existing `_compute_thetas` path and return the same values as `FrequencySpectrum.theta` on both biallelic and multiallelic data, which also makes the previously-unreachable eta branches of `_ac_contribution` and their two normalizer caches live. No existing results change. This fixes a coverage gap identified in #232 

A stale comment above the SFS weight vectors is corrected: it claimed the built-in names go through `_ac_contribution`, but `FrequencySpectrum.theta` uses those weights for built-in names and custom callables alike, while the scalar functions and `diversity_stats` use `_ac_contribution`.

Tests: the four estimators join the implementation-parity harness (scalar via `diversity_stats` versus the SFS dot product), agreeing on every condition including multiallelic; plus absolute values checked against the Achaz definition by hand, and a multiallelic per-allele decomposition check.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -184,6 +184,11 @@ Bug fixes
 * ``mu_ld`` and the ``rogers_huff_r`` family raised a bare
   ``AttributeError``/``TypeError`` on a streaming matrix; they now give
   the same "call ``materialize``" message ``zns``/``omega`` do.
+* ``diversity_stats`` listed the Achaz eta-family estimators (``eta1``,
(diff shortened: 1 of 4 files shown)


andrewkern:
LGTM

andrewkern (approved):

PR #255, opened 2026-09-02
Coverage gap for polyploid `heterozygosity_observed`

nspope:
Cover the ploidy>2 path of heterozygosity_observed, which was untested: an individual is heterozygous when its ploidy haplotypes are not all identical. Cross-check ploidy 3 and 4 against scikit-allel on random data, add a hand-built case that an individual with a missing haplotype is skipped under the default 'include' (and a site with no fully-called individual is NaN), and assert the raise when the haplotype count is not divisible by the ploidy.


nspope:

PR #256, opened 2026-09-02
Cover the untested edges of the SFS helpers

nspope:
Adds coverage for the untested edges of the sfs.py helpers, reusing the module's existing scikit-allel-oracle style: 

- fold_sfs/fold_joint_sfs short-input padding (fold_joint_sfs was previously untested, checked against allel.fold_joint_sfs), 
- joint_sfs returning zeros when no allele segregates in the cohort, 
- project_joint_sfs raising on an up-projection request and returning zeros when nothing survives the filter, 
- and _projection_matrix_vec rejecting n_to > n_from and putting all mass in the empty-sample bin at n_to == 0.
 
Tests only, no code change; sfs.py coverage rises from 71% to 92% on the scoped run.


nspope:

Issue #261, opened 2026-09-03
Read the call_genotype codec once on ZarrGenotypeSource instead of re-probing on disk


andrewkern:
Follow-up from the #223 fix (branch `fix/kvikio-v2-codec-detection`).

`_store_call_genotype_codec` reads the codec by hand-parsing the store's on-disk metadata (`call_genotype/zarr.json` for zarr v3, `call_genotype/.zarray` for v2). That works, but it sits below the layer where the store is already open, and it makes pg_gpu encode zarr's on-disk file schema itself.

`ZarrGenotypeSource.__init__` (`pg_gpu/zarr_source.py`) already opens the store and holds `cg = self._store["call_genotype"]`, then reads `self.chunks = tuple(cg.chunks)`. The codec is available off that same handle via `cg.metadata.to_dict()`:

- v3 -> `{"codecs": [{"name": "zstd"}, ...]}`
- v2 -> `{"compressor": {"id": "blosc"}}`

so the name-extraction logic stays the same; only the source of the dict changes from `json.load` to `to_dict()`, and zarr owns the v2-vs-v3 abstraction.

Separately, the `auto` -> `kvikio` path parses the codec twice today: once in `_pick_chunk_fetcher` (`streaming_matrix.py:113`) and again in `KvikioChunkFetcher.__init__` (`streaming_matrix.py:332`).

## Proposed

- Read the codec name once in `ZarrGenotypeSource.__init__`, beside the `chunks` read, and cache it as `source.codec`.
- Have both call sites read `source.codec`.
- Delete `_store_call_genotype_codec` (and the test-only `_store_call_genotype_chunks`).
- Keep the codec name general on the source; the nvCOMP-support decision stays in the fetcher-selection layer.

## Scope notes

- Touches shared streaming infrastructure, so run the full streaming suite (GRM / IBS / LD / local PCA / biobank), not just `test_kvikio_backend.py`.
- The two probe unit tests move from a bare path to a source; the missing-store case then raises at construction instead of the probe returning None, so that test changes shape.
- `cg.metadata.to_dict()` also reads a sharded-v3 store correctly, where the current hand-parse would miss compression nested in the sharding codec config.

PR #262, opened 2026-09-03
Read zarr v2 compressor metadata in the kvikio codec probe

### LINKED ISSUE #223: kvikio backend cannot engage on zarr v2 stores and misdiagnoses why
Found in the release audit's scale stress run.

`_store_call_genotype_codec` reads only `call_genotype/zarr.json`, which is zarr v3 metadata. On a v2 store (`.zarray` - the bio2zarr default, including the ag1000g store) the codec probe returns None, so:

- `backend='auto'` silently routes to the host fetcher with no hint that GPU-direct reads were skipped;
- `backend='kvikio'` raises "requires a call_genotype codec the nvCOMP GPU decoder supports (['blosc','deflate','lz4','zstd']); could not identify one" - on a store whose codec (blosc) is in that list. The message blames the codec; the cause is v2 metadata never being read;
- the `BadlyChunkedWarning` re-encode hint is unreachable on any v2 store, since it gates on a non-None codec.

Fix: read the v2 `.zarray` compressor metadata, or state plainly that kvikio needs zarr v3. (Separate, by-design gate on the ag1000g store: its chunks span the full sample axis, which routes to host regardless.)

andrewkern:
## Summary

`_store_call_genotype_codec` read only `call_genotype/zarr.json`, the zarr v3 metadata file. A zarr v2 store keeps its metadata in `.zarray` (the bio2zarr default, including the ag1000g store), so the probe found no file and returned `None`. That routed `backend='auto'` to the host fetcher with no hint why, made `backend='kvikio'` raise a misleading "could not identify a codec" error, and left the `BadlyChunkedWarning` re-encode hint unreachable on v2 stores -- even though the store's blosc compressor is one nvCOMP decodes.

The probe now reads whichever metadata layout is present: the codec pipeline in a v3 `zarr.json`, or the single `compressor` in a v2 `.zarray`.

## Tests

- `test_codec_probe_reads_v2_store` -- the probe returns a supported codec on a zarr v2 store.
- `test_auto_picks_kvikio_on_v2_store` -- a v2 store with bio2zarr-shaped chunks routes to kvikio (regression: it fell back to host before).
- `test_explicit_kvikio_on_v2_store_does_not_raise` -- `backend='kvikio'` builds the fetcher instead of raising.

The `_write_vcz_with_sample_chunk` helper gained a `zarr_format` param so one helper writes both layouts. Full `test_kvikio_backend.py` suite passes (16/16).

## Follow-up

#261 tracks a deeper refactor (read the codec once on `ZarrGenotypeSource`, drop the standalone probes) that touches shared streaming infrastructure, kept separate from this fix.

Closes #223.

### DIFF
--- a/pg_gpu/streaming_matrix.py
+++ b/pg_gpu/streaming_matrix.py
@@ -141,32 +141,41 @@ def _pick_chunk_fetcher(source, *, backend):
 
 
 def _store_call_genotype_codec(store_path):
-    """Return the name of the bytes-encoder codec on the store's
+    """Return the name of the compression codec on the store's
     call_genotype array (e.g. ``'zstd'``), or None when the codec
-    spec cannot be read.
+    cannot be read or is not one nvCOMP decodes on the GPU.
 
-    The codec name is what matters for picking a fetcher: nvCOMP has
-    GPU decoders for a fixed list, and the rest go through the host
-    path. Reading the spec is cheap (one JSON parse off the
-    ``call_genotype/zarr.json`` blob).
+    The codec name is what picks a fetcher: nvCOMP has GPU decoders
+    for a fixed list, and the rest go through the host path. The
+    metadata layout differs by zarr format, so read whichever is
+    present: a zarr v3 store keeps a codec pipeline in ``zarr.json``,
+    while a zarr v2 store (the bio2zarr default) keeps a single
+    ``compressor`` in ``.zarray``.
     """
     import json
     import os
-    spec_path = os.path.join(str(store_path), "call_genotype", "zarr.json")
-    if not os.path.exists(spec_path):
-        return None
+    array_dir = os.path.join(str(store_path), "call_genotype")
+    v3_spec = os.path.join(array_dir, "zarr.json")
+    v2_spec = os.path.join(array_dir, ".zarray")
     try:
-        with open(spec_path) as f:
-            spec = json.load(f)
-        for entry in spec.get("codecs", []):
-            name = entry.get("name") if isinstance(entry, dict) else None
-            if name in _NVCOMP_SUPPORTED_CODECS:
-                return name
-            if name == "bytes":
-                continue  # transparent reinterpretation; not a real codec
+        if os.path.exists(v3_spec):
+            with open(v3_spec) as f:
+                spec = json.load(f)
+            names = [entry.get("name") for entry in spec.get("codecs", [])
+                     if isinstance(entry, dict)]
+        elif os.path.exists(v2_spec):
+            with open(v2_spec) as f:
+                spec = json.load(f)
+            compressor = spec.get("compressor") or {}
+            names = [compressor.get("id")]
+        else:
+            return None
+    except (OSError, json.JSONDecodeError):
         return None
-    except (OSError, json.JSONDecodeError, KeyError):
-        return None
+    for name in names:
+        if name in _NVCOMP_SUPPORTED_CODECS:
+            return name
+    return None
 
 
 def _store_call_genotype_chunks(store_path):
--- a/tests/test_kvikio_backend.py
+++ b/tests/test_kvikio_backend.py
@@ -15,18 +15,20 @@ import zarr
 from pg_gpu import HaplotypeMatrix
 from pg_gpu.streaming_matrix import (
     BadlyChunkedWarning, HostChunkFetcher, KvikioChunkFetcher,
-    _pick_chunk_fetcher, _store_call_genotype_chunks,
-    _store_call_genotype_codec,
+    _NVCOMP_SUPPORTED_CODECS, _pick_chunk_fetcher,
+    _store_call_genotype_chunks, _store_call_genotype_codec,
 )
 from pg_gpu.zarr_source import ZarrGenotypeSource
 
 from .conftest import simulate_hm
 
 
-def _write_vcz_with_sample_chunk(path, hm, *, sample_chunk):
+def _write_vcz_with_sample_chunk(path, hm, *, sample_chunk, zarr_format=3):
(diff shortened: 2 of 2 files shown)


pg_gpu/streaming_matrix.py
```diff
@@ -141,32 +141,41 @@ def _pick_chunk_fetcher(source, *, backend):
...
-            if name == "bytes":
-                continue  # transparent reinterpretation; not a real codec
-        return None
-    except (OSError, json.JSONDecodeError, KeyError):
+        if os.path.exists(v3_spec):
+            with open(v3_spec) as f:
+                spec = json.load(f)
+            names = [entry.get("name") for entry in spec.get("codecs", [])
+                     if isinstance(entry, dict)]
+        elif os.path.exists(v2_spec):
```
nspope, line 166: getting a crash here when passing v2 stores to kvikio, because the host-only `numcodecs` is handed a cupy chunk-buffer and can't copy it.

```python
import os, shutil, zarr, numpy as np, cupy as cp
from pg_gpu.zarr_io import write_vcz
from pg_gpu.zarr_source import ZarrGenotypeSource
from pg_gpu.streaming_matrix import (_pick_chunk_fetcher, KvikioChunkFetcher,
                                     HostChunkFetcher, _store_call_genotype_codec)
from tests.conftest import simulate_hm

path = "/tmp/bio2zarr_v2_repro.vcz"
if os.path.exists(path):
    shutil.rmtree(path)
hm = simulate_hm()
haps = hm.haplotypes.get() if hasattr(hm.haplotypes, "get") else hm.haplotypes
gt = HaplotypeMatrix._haplotypes_to_gt(haps)
pos = hm.positions.get() if hasattr(hm.positions, "get") else hm.positions
samples = [f"s{i}" for i in range(gt.shape[1])]
with zarr.config.set({"default_zarr_format": 2}):          # bio2zarr default
    write_vcz(path, gt, pos, samples=samples, contig_name="1",
              chunks=(min(gt.shape[0], 10_000), max(1, gt.shape[1] // 4), 2))

src = ZarrGenotypeSource(path)
print("codec probe:", _store_call_genotype_codec(path))     # -> 'blosc'
f = _pick_chunk_fetcher(src, backend="auto")
print("fetcher:", type(f).__name__)                         # -> KvikioChunkFetcher
try:
    got = list(f.iter_chunks([(0, 10_000), (10_000, 20_000)], prefetch=0))
    host = list(HostChunkFetcher(src).iter_chunks([(0, 10_000), (10_000, 20_000)],
                                                  prefetch=0))
    ok = all(np.array_equal(cp.asnumpy(a[3]), np.asarray(b[3]))
             for a, b in zip(got, host))
    print("matches host:", ok)
except Exception as e:
    print(f"iter_chunks CRASH: {type(e).__name__}: {e}")
finally:
    f.close()

# Observed:
#   codec probe: blosc
#   fetcher: KvikioChunkFetcher
#   iter_chunks CRASH: ValueError: Unable to avoid copy while creating an array
#                      as requested.
```

tests/test_kvikio_backend.py
```diff
@@ -105,6 +130,24 @@ def test_explicit_host_skips_kvikio_path(self, bio2zarr_store):
...
 
+    def test_auto_picks_kvikio_on_v2_store(self, bio2zarr_v2_store):
+        # Regression: a v2 store with a supported compressor and
+        # bio2zarr-shaped chunks must route to kvikio. Before the v2
+        # metadata read the codec probe returned None and auto fell
+        # back to the host fetcher with no hint why.
+        source = ZarrGenotypeSource(bio2zarr_v2_store)
+        fetcher = _pick_chunk_fetcher(source, backend="auto")
+        assert isinstance(fetcher, KvikioChunkFetcher)
+        fetcher.close()
```
nspope, line 141: I think we'd need to iterate a chunk here to actually trigger the crash mentioned in comment above

pg_gpu/streaming_matrix.py
```diff
@@ -141,32 +141,41 @@ def _pick_chunk_fetcher(source, *, backend):
...
+            with open(v2_spec) as f:
+                spec = json.load(f)
+            compressor = spec.get("compressor") or {}
```
nspope, line 169: If `compressor` is a bare string then this'll raise an attribute error

```python
import os, json, numcodecs
from pg_gpu.streaming_matrix import _store_call_genotype_codec, _NVCOMP_SUPPORTED_CODECS

print(sorted(_NVCOMP_SUPPORTED_CODECS))  # ['blosc', 'deflate', 'lz4', 'zstd']

# a non-dict compressor escapes the except clause
d = "/tmp/fake_v2.vcz/call_genotype"
os.makedirs(d, exist_ok=True)
with open(os.path.join(d, ".zarray"), "w") as fh:
    json.dump({"zarr_format": 2, "compressor": "zstd"}, fh)   # compressor is a string
_store_call_genotype_codec("/tmp/fake_v2.vcz")
# AttributeError: 'str' object has no attribute 'get'
```

nspope (commented):
I don't have my head around the constraints here, but asking claude to probe turned up a couple issues

andrewkern:
okay these are really helpful. thanks @nspope. I was working on this late yesterday and was tired, and towards the end realized that the right way to fix this might be #261 -- maybe take a look at that idea if you have a moment?

nspope:
I looked at #261 and it's pretty clearly an improvement---I'd supercede this PR with it. But, the same "chunk-buffer passed to decompressor" crash bug that I posted above is still going to happen, AFAICT, because the issue is a cupy GPU array going to CPU numcodecs for decode. So that's a separate issue that'll still need to be patched to support both v2 and v3 in practice

andrewkern:
okay i'll work on this and close down this PR

PR #265, opened 2026-09-03
Route diploid/dosage `population=` through `get_population_matrix`

### LINKED ISSUE #263: `population=` list argument crashes (and skips validation) on the diploid/dosage paths
Followup deferred from fixing #234.

The single-population statistics document `population` as `str or list` and route it through the shared `get_population_matrix` (`pg_gpu/_utils.py`), which accepts either form and validates a row list with `check_sample_set_rows`. Four diploid/dosage entry points instead resolve `population` by hand and diverge from that contract. All four were confirmed by running against a small `GenotypeMatrix` with `sample_sets={'p1': [0, 1], 'p2': [2, 3]}`.

- `diversity.diplotype_frequency_spectrum` (`diversity.py:1602`) does `genotype_matrix.sample_sets.get(population)`, so a list raises `TypeError: unhashable type: 'list'`. `diplotype_frequency_spectrum(gm, population=[0, 1])` crashes; the name form works.
- `diversity._daf_histogram_diploid` (`diversity.py:1653`) has the same pattern, reached through `daf_histogram(gm, population=[0, 1])` on a `GenotypeMatrix`. Same `TypeError`.
- `selection.garud_h` on a `GenotypeMatrix` dispatches to `_garud_h_diploid` (`selection.py:279`) which forwards `population` into `diplotype_frequency_spectrum`, inheriting the crash. `garud_h(gm, population=[0, 1])` raises while `garud_h(hm, population=[0, 1, 2, 3])` on a `HaplotypeMatrix` works, so one documented API behaves differently by matrix type.
- `decomposition._prepare_dosage` (`decomposition.py:308`, reached via `pca_dosage`/`pca_genotype`) takes the `else: idx = list(population); geno = geno[idx, :]` branch without `check_sample_set_rows`. CuPy fancy indexing is unchecked, so `pca_dosage(gm, population=[0, 99])` on a 4-individual matrix returns a result with no error (out-of-range read / silent garbage) instead of raising, and an empty list hits a low-level reduction error rather than a clear message.

Root cause: these paths hand-roll population resolution instead of using `get_population_matrix`. Routing all four through it fixes the list-argument crashes and adds the missing row validation in one place. The name form and the unknown-population raises are already correct, so this is scoped to the list argument and the dosage validation.

Coverage note: `tests/test_population_coverage.py` exercises the diploid paths only with a name and the not-found case, so it does not catch these; the one list-argument test (`divergence.dxy`) targets `_get_population_indices`, which is already correct. A fix should add a list-argument case (equal to the name form) and an out-of-range/empty case (raises) for the diploid and dosage paths.

nspope:
Closes #263

### DIFF
--- a/pg_gpu/decomposition.py
+++ b/pg_gpu/decomposition.py
@@ -296,18 +296,9 @@ def _prepare_dosage(genotype_matrix, population=None, missing_data='include'):
     if matrix.device == 'CPU':
         matrix.transfer_to_gpu()
 
-    geno = matrix.genotypes
-    # Subset individuals inline: the shared get_population_matrix speaks
-    # haplotype rows, not individual rows.
     if population is not None:
-        if isinstance(population, str):
-            if matrix.sample_sets is None or population not in matrix.sample_sets:
-                raise ValueError(
-                    f"Population {population} not found in sample_sets")
-            idx = matrix.sample_sets[population]
-        else:
-            idx = list(population)
-        geno = geno[idx, :]
+        matrix = _get_population_matrix(matrix, population)
+    geno = matrix.genotypes
 
     n_ind, n_var = geno.shape
     if n_var and int(geno.max()) > 2:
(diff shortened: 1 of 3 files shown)


nspope:
This one is eensy and adds no new machinery so I'm merge

PR #271, opened 2026-09-04
Route zarr v2 stores to the host fetcher; read the codec on the source

### LINKED ISSUE #223: kvikio backend cannot engage on zarr v2 stores and misdiagnoses why

### LINKED ISSUE #261: Read the call_genotype codec once on ZarrGenotypeSource instead of re-probing on disk

andrewkern:
Supersedes #262 with the correct fix for #223, and folds in the #261 refactor.

## The bug (caught by @nspope on #262)

kvikio's nvCOMP GPU decoder works only through zarr v3's codec pipeline. A zarr v2 store -- the bio2zarr default, including the ag1000g store -- decodes through CPU `numcodecs`. #262 made the codec probe *detect* the v2 blosc compressor and route the store to kvikio, which then handed a cupy chunk buffer to `numcodecs` and crashed on the first chunk read:

```
ValueError: Unable to avoid copy while creating an array as requested.
```

Detecting the v2 codec cannot help: v2 fundamentally cannot be GPU-decoded through zarr's pipeline.

## The fix

Gate kvikio on **a zarr v3 store with an nvCOMP-supported codec** -- the true precondition for GPU decode.

- `backend='auto'` on a v2 store -> host fetcher, with a size-gated `KvikioUnavailableWarning` hint (silent on small stores).
- `backend='kvikio'` on a v2 store -> clear `ValueError`: "kvikio GPU decode requires a zarr v3 store ...", instead of a misleading "could not identify a codec" or a mid-read crash.

## Folded-in refactor (#261)

Read the codec name and zarr format **once** on `ZarrGenotypeSource` (`source.codec`, `source.zarr_format`), off the `call_genotype` handle `__init__` already opens, and delete the two standalone on-disk probe functions. Reading through zarr's metadata API also removes the hand-parsing of the v2/v3 on-disk layout, so the malformed-`compressor` `AttributeError` @nspope flagged can no longer arise.

## Tests

- `test_v2_store_reports_compressor`, `test_v3_store_reports_zstd` -- source reads codec/format for both layouts.
- `test_explicit_kvikio_on_v2_store_raises` -- v2 + `backend='kvikio'` raises "needs v3".
- `test_v2_store_host_iterates_without_crash` -- the crash regression: a v2 store routes to host and iterates a chunk cleanly (@nspope's point that the old tests never iterated).
- Full streaming suite (GRM / IBS / LD / local PCA / biobank / kvikio): 207 passed on an A100.

Closes #223. Closes #261.

### DIFF
--- a/pg_gpu/streaming_matrix.py
+++ b/pg_gpu/streaming_matrix.py
@@ -122,8 +122,8 @@ def _pick_chunk_fetcher(source, *, backend):
                 f"{source.path}: zarr v{source.zarr_format} store; the "
                 f"kvikio GPU decoder needs zarr v3 (a v2 compressor decodes "
                 f"on the CPU). Falling back to the host fetcher. Re-encode "
-                f"as zarr v3 via HaplotypeMatrix.vcf_to_zarr to enable "
-                f"kvikio.",
+                f"as zarr v3 with HaplotypeMatrix.vcf_to_zarr(..., "
+                f"zarr_format=3) to enable kvikio.",
                 KvikioUnavailableWarning, stacklevel=4,
             )
         return HostChunkFetcher(source)
@@ -143,8 +143,9 @@ def _pick_chunk_fetcher(source, *, backend):
                 f"full sample axis ({source.num_diploids} diploids); the "
                 f"kvikio fetcher gives no speedup at this chunking. "
                 f"Falling back to the host fetcher. Re-encode with "
-                f"bio2zarr-style chunking (sample_chunk ~= 1000) via "
-                f"HaplotypeMatrix.vcf_to_zarr to enable kvikio.",
+                f"HaplotypeMatrix.vcf_to_zarr(..., zarr_format=3), whose "
+                f"bio2zarr sample chunking splits the sample axis, to "
+                f"enable kvikio.",
                 BadlyChunkedWarning, stacklevel=4,
             )
         return HostChunkFetcher(source)
@@ -309,8 +310,8 @@ class KvikioChunkFetcher(ChunkFetcher):
                 f"kvikio GPU decode requires a zarr v3 store; "
                 f"{source.path} is zarr v{source.zarr_format}, whose "
                 f"numcodecs compressor decodes on the CPU. Re-encode as "
-                f"zarr v3 via HaplotypeMatrix.vcf_to_zarr to use the "
-                f"kvikio backend."
+                f"zarr v3 with HaplotypeMatrix.vcf_to_zarr(..., "
+                f"zarr_format=3) to use the kvikio backend."
             )
         if source.codec not in _NVCOMP_SUPPORTED_CODECS:
             raise ValueError(
--- a/pg_gpu/streaming_matrix.py
+++ b/pg_gpu/streaming_matrix.py
@@ -32,7 +32,7 @@ import zarr
 from kvikio.zarr import GDSStore
 
 from ._gpu_genotype_prep import build_genotype_matrix, build_haplotype_matrix
-from ._warnings import BadlyChunkedWarning
+from ._warnings import BadlyChunkedWarning, KvikioUnavailableWarning
 
 
 def _pad_to(arr, shape):
@@ -110,9 +110,22 @@ def _pick_chunk_fetcher(source, *, backend):
     if backend == "kvikio":
         return KvikioChunkFetcher(source)
 
-    codec = _store_call_genotype_codec(source.path)
-    if codec is None:
-        # codec unreadable or unsupported -> host
+    if not _kvikio_can_decode(source):
+        # A zarr v2 store decodes through CPU numcodecs, and an
+        # uncompressed or unsupported codec has no nvCOMP GPU decoder --
+        # either way kvikio cannot engage, so use the host fetcher. Hint
+        # on the v2 case (the common, fixable one) when the store is big
+        # enough that the GPU path would have mattered; small fixtures
+        # would just get noise.
+        if source.zarr_format != 3 and _store_large_enough_to_warn(source):
+            warnings.warn(
+                f"{source.path}: zarr v{source.zarr_format} store; the "
+                f"kvikio GPU decoder needs zarr v3 (a v2 compressor decodes "
+                f"on the CPU). Falling back to the host fetcher. Re-encode "
+                f"as zarr v3 via HaplotypeMatrix.vcf_to_zarr to enable "
+                f"kvikio.",
+                KvikioUnavailableWarning, stacklevel=4,
+            )
         return HostChunkFetcher(source)
     chunks = source.chunks  # already cached on the source from __init__
     whole_sample_axis = (chunks is not None and len(chunks) >= 2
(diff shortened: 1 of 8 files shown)


pg_gpu/streaming_matrix.py
```diff
@@ -110,9 +110,22 @@ def _pick_chunk_fetcher(source, *, backend):
...
+        # either way kvikio cannot engage, so use the host fetcher. Hint
+        # on the v2 case (the common, fixable one) when the store is big
+        # enough that the GPU path would have mattered; small fixtures
+        # would just get noise.
+        if source.zarr_format != 3 and _store_large_enough_to_warn(source):
+            warnings.warn(
+                f"{source.path}: zarr v{source.zarr_format} store; the "
+                f"kvikio GPU decoder needs zarr v3 (a v2 compressor decodes "
+                f"on the CPU). Falling back to the host fetcher. Re-encode "
+                f"as zarr v3 via HaplotypeMatrix.vcf_to_zarr to enable "
```
nspope: I may be misunderstanding---but doesn't `vcf_to_zarr` use `bio2zarr`, which defaults to output v2 (see: https://github.com/sgkit-dev/bio2zarr/blob/053832006f43d2f0125891b7fbd624757f733098/bio2zarr/zarr_utils.py#L22-L25)? So this advice wouldn't work. Rather, we'd need to wire through a `zarr_format=3` through `vcf_to_zarr` to `bio2zarr.encode`, here: https://github.com/kr-colab/pg_gpu/blob/8ce9aee3e56ada5566a5cad04da479b5cfeca8ac/pg_gpu/zarr_io.py#L726

and/or use `HaplotypeMatrix.to_zarr` which seems more natural

pg_gpu/streaming_matrix.py
```diff
@@ -320,16 +304,22 @@ def __init__(self, source, *, compat_mode="ON", num_threads=8):
...
                 f"got {compat_mode!r}"
             )
-        codec = _store_call_genotype_codec(source.path)
-        if codec is None:
+        if source.zarr_format != 3:
+            raise ValueError(
+                f"kvikio GPU decode requires a zarr v3 store; "
+                f"{source.path} is zarr v{source.zarr_format}, whose "
+                f"numcodecs compressor decodes on the CPU. Re-encode as "
+                f"zarr v3 via HaplotypeMatrix.vcf_to_zarr to use the "
```
nspope: same comment as above re bio2zarr defaulting to v2

pg_gpu/streaming_matrix.py
nspope: same comment as above re: bio2zarr defaulting to v2

pg_gpu/zarr_source.py
```diff
@@ -31,6 +31,29 @@
...
+    compressor, so the caller never parses the on-disk metadata layout
+    itself. The v3 ``bytes`` codec is a transparent reinterpretation,
+    not a compressor, so it is skipped.
+    """
+    codecs = meta.get("codecs")
+    if codecs is not None:  # zarr v3 codec pipeline
+        for entry in codecs:
+            name = entry.get("name") if isinstance(entry, dict) else None
+            if name and name != "bytes":
+                return name
```
nspope, line 57: this won't necessarily return the right thing; for example if the v3 zarr store is sharded, then the codec name is a level down:

```python
import zarr
from zarr.codecs import ZstdCodec
from pg_gpu.zarr_source import _codec_name_from_metadata
from pg_gpu.streaming_matrix import _NVCOMP_SUPPORTED_CODECS

g = zarr.create_group(store=zarr.storage.MemoryStore(), zarr_format=3)
a = g.create_array(
  "x", shape=(100,), chunks=(100,), shards=(100,),
  dtype="int8", compressors=[ZstdCodec()]
)   # zstd nested in the shard
codec = _codec_name_from_metadata(a.metadata.to_dict())        # zarr's real output
# the nested zstd is missed by `_codec_name...` so the store is treated as non-GPU-decodable.
assert codec not in _NVCOMP_SUPPORTED_CODECS
```

not a big deal, but probably not intended? maybe worth pinning as an xfail (and making last assert fail by removing the not)

nspope (commented):
Looks good, couple small comments

pg_gpu/streaming_matrix.py
andrewkern: Seems easy enough. Will do

pg_gpu/zarr_source.py
andrewkern: Let me add a recurse to pull this and do the right thing

andrewkern:
okay this should be ready @nspope. want to take a quick look? 

I added a `zarr_format` arg to `vcf_to_zarr` (passed through to `bio2zarr.encode`) and pointed the fallback messages at `vcf_to_zarr(..., zarr_format=3)` since bio2zarr defaults to v2, and made the codec probe recurse into sharded v3 stores.

nspope:
Looks good to me

PR #277, opened 2026-09-06
Cover trivial edge branches (dist_moments, moving-window, grm, pc_dist/corners)

andrewkern:
First slice of #238, taking the pure-logic coverage gaps that need no fixtures or special environment. Each added test was confirmed to close its target branch.

- **distance_stats.dist_moments** — the `n<2`, zero-variance (`m2==0`), and `n<4` early returns (small pairwise-distance vectors).
- **resampling._moving_nansum / _moving_nanmean** — windowed cumsum over non-overlapping, strided, NaN, all-NaN, and too-large (empty-result) windows (`resampling.py` now 100%).
- **relatedness.grm** — an all-monomorphic matrix returns a zero GRM instead of dividing by zero.
- **decomposition.pc_dist / corners** — the malformed-input raises: missing `npc` on a flat matrix, unknown `normalize`, non-`(n,2)` input, and too few non-NaN points.

Tests only, no source changes. 89 passed on an A100.

Note: #238 is a multi-module coverage epic; this PR does the trivial tier. The heavier items (streaming-relatedness fixtures, non-VCZ zarr handler, `moments_ld` arg-validation + missing-data comparison, `genotype_matrix` mask/transfer lifecycle, plotting/accessible edge inputs) remain.

Refs #238.

Issue #238, opened 2026-08-25
Coverage gap: matrix / I/O / infrastructure modules


nspope:
Smaller residual gaps across the matrix, decomposition, relatedness, and I/O
modules. Grouped because each is a handful of lines; split into per-module PRs if
preferred. (Baseline: `pixi run coverage` on `main` b0c9801; line numbers drift —
match on function name. `population=` and `missing_data='exclude'` gaps for these
modules are tracked in #233 and #234.)

## genotype_matrix.py (86%, 60/425)
| File:line | Function | Uncovered behavior | Test |
|---|---|---|---|
| 161, 221, 229-231 | accessible-mask setter, `set_/remove_accessible_mask` | mask attach/remove lifecycle | attach then remove a BED mask |
| 504-507, 545, 550 | `from_zarr` | `streaming='always'`+`fields` NotImplementedError; auto/never streaming with fields | those arg combos |
| 276, 285, 295 | `transfer_to_gpu`/`transfer_to_cpu` | direct transfer calls | round-trip a matrix |
| 712-763 | `filter` | shape-mismatch raises + empty-result-after-filter path | bad mask shape; filter-all |
| 817, 841, 869-870 | `load_pop_file` | no-sample-names raise; pop dict/file parse; dropped-pop warning | pop file w/ unknown sample |
| 69-80, 97-105 | `__init__` | empty-input validation; accessible-mask resolution w/o chrom bounds | empty args; mask w/o bounds |
| 578, 635-636 | `_build_eager` | non-VCZ Zarr store exception handler | (hard; needs a non-VCZ zarr) |

## decomposition.py (92%, 60/730) — residual (population subsetting is its own issue)
| File:line | Function | Uncovered behavior | Test |
|---|---|---|---|
| 865, 1169 | `_local_pca_streaming`, `_local_pca_with_jackknife` | empty WindowIterator (`window_size > n_variants`) | window bigger than data |
| 832-837, 1450-1452 | `_local_pca_streaming`, `pc_dist` | NaN rows when `n_variants < max(k,2)`; NaN eigvals | tiny windows |
| 1329 | `_local_pca_with_jackknife` | `aggregate='unknown'` raise | bad aggregate arg |
| 1388, 1396, 1413 | `pc_dist` | `npc=None` on flat matrix; shape mismatch; `normalize='unknown'` | malformed inputs |
[...]

PR #281, opened 2026-09-07
Cover GenotypeMatrix edge branches (init, transfer, filter, pop file, mask)

andrewkern:
Next chunk of the #238 coverage campaign (after #277), taking the `genotype_matrix.py` in-memory public API. Every added test drives one branch directly; each was confirmed to close its target lines under `--cov=pg_gpu.genotype_matrix`.

- **`__init__`** -- empty-genotypes and empty-positions validation raises.
- **`transfer_to_gpu` / `transfer_to_cpu`** -- the CPU<->GPU round-trip, the already-on-device no-op, and carrying the accessible index across devices.
- **`filter`** -- the variants and genotypes mask shape-mismatch raises, and the drop-everything empty-result path.
- **`load_pop_file`** -- the no-sample-names raise, dict and file assignment, and the dropped-empty-population warning.
- **accessible-mask lifecycle** -- `set_accessible_mask` (full and partial masks), `remove_accessible_mask`, and the setter's accessible-index bookkeeping.

Tests only, no source changes. 14 passed on an A100 (most are CPU-only; the transfer cases need the GPU).

Deferred to a later chunk (heavier setup, out of scope here): `from_zarr` with `fields` (needs a fields-bearing zarr store) and the `_build_eager` non-VCZ handler (the issue marks it hard).

Refs #238.

### DIFF
--- a/tests/test_genotype_matrix_coverage.py
+++ b/tests/test_genotype_matrix_coverage.py
@@ -1,0 +1,138 @@
+"""Edge-case coverage for GenotypeMatrix: input validation, device transfer,
+filtering, population-file loading, and the accessible-mask lifecycle.
+
+These are public paths on the core diploid data structure that the broader
+suite exercises only incidentally; each test drives one branch directly.
+"""
+import numpy as np
+import cupy as cp
+import pytest
+
+from pg_gpu import GenotypeMatrix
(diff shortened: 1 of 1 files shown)


nspope (approved):
I think this is good to go

PR #282, opened 2026-09-09
Cover decomposition edge branches (dosage validation, tiny window, jackknife aggregate)

andrewkern:
Next chunk of the #238 coverage campaign (after #277 and #281), finishing the self-contained `decomposition.py` edge cases. Each test drives one branch directly and was verified against the code.

- **`_prepare_dosage`** -- the non-biallelic guard (a dosage > 2 raises) and the empty-after-`exclude` path (every site missing, so the standardized matrix comes back with zero variants instead of dividing by zero).
- **`local_pca`** -- a window with fewer variants than the requested PCs (`n_variants < max(k, 2)`) returns NaN eigenvalue rows rather than erroring.
- **`local_pca_jackknife`** -- an unknown `aggregate` argument raises.

Tests only, no source changes. 45 tests pass across the two touched files on an A100.

Deferred (needs more setup, out of scope here): the empty-`WindowIterator` raise (snp/bp windows always yield at least one window on real data, so triggering zero windows needs a crafted iterator), and the streaming local-PCA path (`_local_pca_streaming`, needs a streaming-matrix fixture).

Refs #238.

### DIFF
--- a/tests/test_local_pca.py
+++ b/tests/test_local_pca.py
@@ -629,3 +629,18 @@ class TestPcDistCornersEdges:
         # Fewer non-NaN points than corners requested cannot be placed.
         with pytest.raises(ValueError, match="at least"):
             corners(np.array([[0.0, 0.0], [1.0, 1.0]]), prop=0.5)
+
+
+class TestLocalPCAEdges:
+    """Degenerate-window and argument-validation guards on local PCA."""
+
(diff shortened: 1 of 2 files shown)


andrewkern:
tiny one here @nspope

nspope (approved):

nspope:
Looks good, merged

PR #283, opened 2026-09-09
Cover streaming relatedness degenerate branches (empty, monomorphic, indexes)

andrewkern:
Next chunk of the #238 coverage campaign (after #277, #281, #282), taking the streaming relatedness path. The existing parity tests use polymorphic, non-empty stores, so these drive the degenerate branches they never reach. Each was verified against the code.

- **`grm`** -- a monomorphic streaming source (no polymorphic site) and an empty streaming source (a region with no variants) both return a zero matrix.
- **`ibs`** -- an empty streaming source returns the identity-diagonal convention, and the population fallback (`_stream_n_ind`) counts individuals from the named sample set.
- **`genetic_relatedness`** -- the `indexes=` argument projects the requested `(i, j)` pairs out of the streaming result.

Tests only, no source changes. 21 pass across this file and the two existing streaming-relatedness files (no regression) on an A100.

Refs #238.

### DIFF
--- a/tests/test_relatedness_streaming_coverage.py
+++ b/tests/test_relatedness_streaming_coverage.py
@@ -1,0 +1,87 @@
+"""Degenerate-source coverage for the streaming relatedness path.
+
+grm / ibs / genetic_relatedness route a streaming matrix through a separate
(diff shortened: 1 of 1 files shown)


nspope:
Looks good

PR #297, opened 2026-09-10
Cover the remaining moments_ld branches

andrewkern:
Covers the `moments_ld.py` branches still open under #238 after #284 (moments env; line numbers on current main):

- `_compute_ld_sums` 365: pairs binned by genetic distance. `test_recombination_binned_run` used `r_bins` up to `5e-5` on a `1e-5 M/bp` map, which caps pairs at 5 bp on a matrix with variants 100 bp apart, so no pair was ever binned and the test passed on an empty loop. The bins now reach `1e-2` (pairs up to about 1 kb), with the inner edge placed between the pair distances so no pair sits on a boundary, and every LD bin is asserted to carry signal. The same run checks that `report=False` prints nothing.
- `compute_ld_statistics` 175, 196: the `report=True` "Loading ..." lines on both VCF paths, checked through `capsys`.
- `_compute_ld_sums` 312: an individual listed in two populations is counted once in the union biallelic filter. A two-site matrix whose second site is called only in the shared individual forms no pair, so every LD bin stays at zero.

Left alone: 168 (`from_haplotype_matrix` always returns a GPU matrix, so the transfer after it cannot run; a candidate for the dead-code issue), 338 (positions are always cupy by the time `_compute_ld_sums` runs on the public path), and the pre-loaded-matrix device branches the issue already excludes.

Refs #238

### DIFF
--- a/tests/test_moments_ld_coverage.py
+++ b/tests/test_moments_ld_coverage.py
@@ -110,8 +110,9 @@ class TestRecMapHelpers:
 
 
 class TestReportRun:
-    """A full report=True run over each compute path prints progress and
-    returns the moments-format result dict."""
+    """Full runs over each compute path: progress reporting on and off,
+    recombination-distance binning, and the union filter across
+    populations, each returning the moments-format result dict."""
 
     def test_genotype_report_run(self):
         gm = _gm()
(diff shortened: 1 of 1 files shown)


nspope (approved):
Looks good. #238 has a follow up comment about missing moments LD parity testing (which is for correctness, not coverage). I can take that if you'd like?

andrewkern:
sure yeah-- if you take that that would be great.

Issue #301, opened 2026-09-11
Missing-data checks materialize a matrix-sized boolean before reducing


andrewkern:
`selection.garud_h`, `moving_garud_h`, `diversity.haplotype_diversity`, `diversity.haplotype_count`, and `HaplotypeMatrix.count_missing` decide whether a matrix has missing calls with `cp.any(hap < 0)` or `cp.sum(hap < 0, axis=0)`. The comparison materializes a boolean array the size of the matrix before the reduction runs: 10.6 GB for the Ag1000G X arm (2,570 haplotypes x 4.1 M sites), about 32 GB for 3R.

Since the exact haplotype hash landed (#207), the hashing paths make no matrix-sized copy, so this check is now the largest allocation on the scalar paths. Measured on the X arm: the memory pool sits at 11.0 GB after a windowed Garud scan (the int8 matrix alone) and grows to 21.3 GB inside `selection.garud_h(hm)`, all of it the `hap < 0` temporary.

#### Sites

- `pg_gpu/selection.py`: the `missing_data='exclude'` filters in `garud_h` and `moving_garud_h` (`cp.sum(hap < 0, axis=0)`), the `cp.any(hap < 0)` gates in `moving_garud_h` and `_distinct_haplotype_frequencies_missing`, and the exclude filters in the nSL/iHS-family functions (`cp.sum(haplotype_matrix.haplotypes < 0, axis=0)`).
- `pg_gpu/diversity.py`: the exclude filters and `cp.any` gates in `haplotype_diversity` and `haplotype_count`, and the `has_missing` check near the daf/frequency helpers.
- `pg_gpu/haplotype_matrix.py`: `count_missing` sums `is_missing()`, which is the same full-size boolean.

#### Suggested direction

- For the has-missing gate: `hap.min() < 0` is an int8 reduction with no temporary.
- For the per-site missing count: `_memutil.allele_counts` already returns `n_valid` per site from one int8 pass, so `n_hap - n_valid` gives the missing count without a boolean copy; `count_missing` and `is_missing`-based callers can route through it.

#### Expected

None of these functions should allocate more than small per-site or per-haplotype vectors on top of the matrix itself.

PR #302, opened 2026-09-11
Hash haplotypes exactly for Garud H and haplotype counts

### LINKED ISSUE #207: Windowed Garud H requests a 109 GB allocation at cohort-scale haplotype counts
Windowed Garud H OOMs at biobank haplotype counts, identically on main
and the multiallelic trunk:

```python
windowed_analysis(hm, window_size=20_000, step_size=20_000,
                  statistics=['garud_h12'])
# hm: 4,662,820 variants x 2,940 haplotypes (Ag1000G X:1-6 Mb)
# OutOfMemoryError: Out of memory allocating 109,669,526,528 bytes
# (allocated so far: 41,126,072,832 bytes)
```

A 109 GB single allocation on an 80 GB card. The fused Garud kernel
sorts haplotype hashes in shared memory per window, but something on the
host side scales as roughly windows x haplotypes x window-capacity;
2,940 haplotypes at 20 kb windows crosses the cliff.

Not a regression -- byte-identical failure on both branches during the
#184 A/B -- but it means the Garud scan silently excludes cohort-scale
data. A chunked launch over window batches, or a capacity-derived batch
size like the fused engines already use elsewhere, would cap the
footprint.

### LINKED ISSUE #294: Garud H for a window with no variants reports a neighbouring window's value
Found during retrospective for #239

When windowed Garud H statistics are computed over non-overlapping windows, a window that contains no variants is given the value of the next window instead of the value for an empty window. Its `garud_h1`, `garud_h12`, `garud_h123`, `garud_h2h1`, and `haplotype_count` come out wrong. If no window in the scan contains any variant, the same code raises `IndexError` instead of returning results. The code path used for overlapping windows handles empty windows correctly, so a scan with overlapping windows and a scan with non-overlapping windows can disagree on the same data.

#### Root cause

For non-overlapping windows, `_garud_h_per_window_reduceat` in `pg_gpu/windowed_analysis.py` builds each window's per-haplotype hash by summing the contributions of the variants that fall in that window, grouping them with `cp.add.reduceat`. `cp.add.reduceat` does not return zero for an empty group: when a window's start and stop indices are equal (no variants), it returns the first value of the following group instead. So an empty window is assigned the hash contribution of the next window's first variant, and the Garud calculation then describes that neighbour rather than the empty window.

The same routine also slices the variant columns down to the range spanned by the windows before grouping. When no window covers any variant, that range is empty, and `cp.add.reduceat` over an empty set of columns raises `IndexError: cannot do a non-empty take from an empty axes`.

#### Reproduction

A non-overlapping 100 base-pair grid over variants at only positions 10 and 210 leaves the middle window from 100 to 200 with no variants:

```python
import numpy as np
from pg_gpu.windowed_analysis import windowed_analysis

hap = np.array([[0, 1], [1, 0], [1, 1], [0, 0]], dtype=np.int8)
m = HaplotypeMatrix(hap, np.array([10, 210], dtype=np.int64), 0, 300)
df = windowed_analysis(m, window_size=100, step_size=100, statistics=["garud_h1"])
print(df[["start", "end", "n_variants", "garud_h1"]])
```

The window with `n_variants == 0` reports `garud_h1 = 0.5`, borrowed from the variant at position 210, instead of `1.0`. A region whose windows all lie past the last variant raises `IndexError`.

#### Correct behaviour

A window with no variants has no sites that distinguish the haplotypes, so every haplotype is identical: one distinct haplotype at frequency 1. That gives `garud_h1`, `garud_h12`, and `garud_h123` equal to 1, `garud_h2h1` equal to 0, and `haplotype_count` equal to 1. The code path used for overlapping windows, `_garud_h_single_pass`, already produces exactly this (its running-sum difference over an empty window is zero), and the non-overlapping path should match it.

#### Impact

Any windowed scan that steps by the window size (so the windows do not overlap) and contains a window with no variants, such as a gene desert, a masked stretch, or a window past the last variant, gets a wrong Garud H for that window with no error raised. A run of empty windows at the start or end of a region turns into an `IndexError`. Because the overlapping-window path is correct, results also differ between overlapping and non-overlapping scans of the same data.

#### Suggested direction

After the grouping step, set the hash of any window whose start and stop indices are equal to zero, so the Garud kernel treats all haplotypes in an empty window as identical, and handle the no-variant-anywhere case by returning the empty-window value for every window rather than slicing an empty set of columns. This corrects how empty windows are summed and does not change the Garud definition.

#### Status

A strict expected-failure test added in #295, `test_garud_tile_empty_window_value` in `tests/test_windowed_ld_garud_coverage.py`, pins the intended empty-window result and currently fails because of the borrowed value. Once the summing is fixed the test passes, and the strict marker then flags itself for removal.

andrewkern:
Closes #207. Closes #294.

## Problem

- Windowed Garud H on a non-overlapping grid upcast the whole haplotype matrix to float64 three times over (109 GB each for 2,940 haplotypes across 4.7 M sites) with no memory check, and the fused engine held a transposed int8 copy on top.
- `cp.add.reduceat` gave a window with no variants the next window's first-variant hash; a grid lying entirely past the last variant raised `IndexError`. Sliding grids whose overlaps hold no variants took the same path.
- The fused kernel sorted each window in shared memory with a sort that never reached haplotypes past index 2048, so H statistics and `haplotype_count` were silently wrong above 2,048 haplotypes (50 patterns among 2,940 haplotypes came back as 913 distinct) and the kernel failed outright above 3,072.
- `haplotype_diversity`, `haplotype_count` and `garud_h` made a float32 copy of the matrix; `moving_garud_h` made five float64 copies.

## Fix

- New `pg_gpu/_haplotype_hash.py`. One int8-native kernel sums per-variant weights in index order for each (window, haplotype) pair straight from the matrix (any signed integer dtype, C or F order, read in place through byte strides), so identical haplotypes hash bit-identically and an empty window hashes to zero. Each window is sorted in global memory by two stable per-row argsorts, and a walk kernel tallies run lengths as exact integers so the statistics depend only on the run-length multiset. Windows are batched to free GPU memory through `_memutil.estimate_garud_window_batch`.
- `windowed_analysis.py` drops the reduceat, prefix-sum and chunked Garud paths and the old kernel (486 lines fewer). Garud-only requests skip the transposed copy and never take the chunked detour; the eager path reuses the existing allele-cap filter.
- `haplotype_diversity`, `haplotype_count`, `garud_h` and `moving_garud_h` share the same hash. `moving_garud_h` rejects windows outside the matrix. The four H formulas live in one function used by every path.
- Tests: a unit file for the kernels, empty-window coverage on tile and sliding grids, batching invariance, a 2,500-haplotype sort regression, a scikit-allel oracle for `moving_garud_h`, and a slow 6 GB stress case. The strict xfail from #295 is removed.
- Docs, changelog, and the biobank example help text updated.

## Verification

| Check | Result |
|---|---|
| Full suite | 1784 passed, 80 skipped, 25 xfailed, 1 failed (`test_scikit_allel_comparison_runs_small`, a 180 s subprocess timeout on a heavily loaded host; the script does not touch the changed code) |
| Bit-exact eager vs streaming Garud tests | pass |
| Regression diff against results saved from `main` (simulated and `ag1000g.X.2Mb.n113`, tile and sliding grids) | match within 1e-9 |
| Slow stress, 3,000 x 2,000,000 | passes in 24 s; the old path needed 144 GB |
| 2,940 x 500,000 tile scan, 300 windows | 0.15 s vs 0.67 s before, no large intermediates |
| Ag1000G X arm, 2,570 x 4.1 M, 20 kb tiles | pool memory 11 GB (the int8 matrix), 3 empty windows now report one haplotype, spot-checked windows match exact unique rows |

Sliding grids at extreme overlap (window 100 x step) are about 2.5x slower than the old prefix-sum path on the 500k-variant benchmark; at overlap 10 they are faster.

The scalar paths still materialize a matrix-sized boolean in their pre-existing missing-data check; that is #301.

### DIFF
--- a/pg_gpu/_haplotype_hash.py
+++ b/pg_gpu/_haplotype_hash.py
@@ -1,0 +1,283 @@
+"""Exact haplotype-identity hashing shared by Garud's H and haplotype counting.
+
+Each haplotype in a window is reduced to two float64 hashes: the in-order sum
+of a per-variant weight over the alleles it carries. One thread computes one
+haplotype's sum in variant-index order, so two identical haplotypes give
+bit-identical hashes, and a window with no variants hashes to zero. Distinct
+haplotypes differ by a random combination of weights that is never within
+rounding distance of zero, so grouping by exact hash equality counts distinct
+haplotypes exactly. The kernels read the int8 matrix directly; nothing here
+materializes a float copy of the haplotype matrix, and the per-window sort runs
+in global memory, so the haplotype count is unbounded.
+"""
+
+import cupy as cp
+import numpy as np
+
+from . import _memutil
+
+_GARUD_SALT1 = 0x9E3779B97F4A7C15   # golden-ratio constant
+_GARUD_SALT2 = 0xC6BC279692B5C323   # phi^-2-based companion
+
+_THREADS = _memutil._THREADS_PER_BLOCK
+
+# Variants per segment when hashing whole rows. Splitting a long row into
+# segments gives the GPU many threads per haplotype instead of one.
+_ROW_SEGMENT = 4096
+
+
+def _index_weights(n_var, salt):
+    """Float64 weights in ``[-1, 1)`` for variant indices ``0 .. n_var - 1``.
+
+    Splitmix64-scrambles each index with the salt. Two salts give two
+    independent weight columns; together they make a collision between
+    distinct haplotype patterns vanishingly rare. The distribution is
+    uniform rather than normal because the weights only tell patterns
+    apart.
+    """
+    s = cp.uint64(salt & 0xFFFFFFFFFFFFFFFF)
+    x = cp.arange(n_var, dtype=cp.uint64) + s
+    x = (x ^ (x >> cp.uint64(30))) * cp.uint64(0xBF58476D1CE4E5B9)
+    x = (x ^ (x >> cp.uint64(27))) * cp.uint64(0x94D049BB133111EB)
+    x = x ^ (x >> cp.uint64(31))
+    # Mantissa-pack the low 52 bits as a float64 in [1, 2); subtract 1 to
+    # get [0, 1); rescale to [-1, 1).
+    mant = (x & cp.uint64(0x000FFFFFFFFFFFFF)) | cp.uint64(0x3FF0000000000000)
+    u = mant.view(cp.float64) - 1.0
+    return 2.0 * u - 1.0
+
+
+def hash_weights(n_var):
+    """The two weight vectors ``(w1, w2)`` for a matrix with ``n_var`` variants."""
+    return _index_weights(n_var, _GARUD_SALT1), _index_weights(n_var, _GARUD_SALT2)
+
+
+# One thread per (window, haplotype). The matrix is addressed with byte
+# strides through a signed char pointer, so int8 and wider signed integer
+# matrices in C or F order are read in place: a wider element is read through
+# its low byte, which holds the allele index for the small values a haplotype
+# matrix carries (-1 for missing, 0.. for alleles). All offsets are 64-bit.
+_window_hash_kernel = cp.RawKernel(r'''
+extern "C" __global__
+void garud_window_hash(const signed char* hap, long long stride0, long long stride1,
+                       const double* w1, const double* w2,
+                       const long long* win_start, const long long* win_stop,
+                       int n_hap, int n_windows, int blocks_per_window,
+                       double* out1, double* out2) {
+    long long b = blockIdx.x;
+    int w = (int)(b / blocks_per_window);
+    int h = (int)(b % blocks_per_window) * blockDim.x + threadIdx.x;
+    if (w >= n_windows || h >= n_hap) return;
+    const signed char* row = hap + (long long)h * stride0;
+    long long lo = win_start[w];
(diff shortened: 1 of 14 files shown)


pg_gpu/_haplotype_hash.py
```diff
@@ -0,0 +1,283 @@
...
+    out_sum_f2[w] = (double)sum_r2 / (n * n);
+    out_top0[w] = (double)top0 / n;
+    out_top1[w] = (double)top1 / n;
+    out_top2[w] = (double)top2 / n;
+    out_n_distinct[w] = (double)n_distinct;
+}
+''', 'garud_walk')
+
+
+def garud_from_moments(sum_f2, top0, top1, top2):
```
nspope, line 136: is there any risk of divide-by-zero here? say, for empty windows or sample sets?

nspope (approved):
Looks great. There's a small pre-existing consistency issue (#303) to fix at some point, but that's orthogonal to the fixes here.

pg_gpu/_haplotype_hash.py
andrewkern: nope that should be totally safe. No variants does the right thing and the sample set being empty is covered by `check_sample_set_rows`

PR #307, opened 2026-09-15
Preserve metadata fields in subset

### LINKED ISSUE #181: Variant-subset methods drop sample names and QC fields
Came up during #180 (which adds `restrict_to_biallelic` / `restrict_to_segregating`). These were found to drop `samples` and `fields` while subsetting the data objects. An audit shows the same metadata loss across the whole variant-subset family, not just those methods.

#### Behaviour

`HaplotypeMatrix` and `GenotypeMatrix` carry two pieces of metadata beyond the genotype array and positions:

- `samples`: per-sample names. These are per-sample, so a *variant* subset does not change them. `load_pop_file` raises `"No sample names stored"` when it is `None`.
- `fields`: per-variant `(n_var,)` or per-genotype `(n_var, n_samples)` QC arrays (host numpy), loaded via `from_vcf(fields=...)` or from zarr. They are variant-indexed, so a variant subset must slice them on axis 0.

Several methods that derive a new matrix by subsetting variants drop one or both instead of carrying/slicing them. As a result `mat.restrict_to_biallelic().load_pop_file(...)` raises, and QC fields loaded at parse time silently vanish after any of these calls.

Affected public methods:

| Method | drops `samples` | drops `fields` |
|---|---|---|
| `HaplotypeMatrix.get_subset` | yes | yes |
| `HaplotypeMatrix.get_subset_from_range` | yes | yes |
| `HaplotypeMatrix.restrict_to_biallelic` | yes | yes |
| `HaplotypeMatrix.restrict_to_segregating` | yes | yes |
| `GenotypeMatrix.restrict_to_segregating` | no | yes |

`HaplotypeMatrix.get_subset`'s empty-result branch additionally never sets `.fields` on the returned object, so `.fields` access on a fully-filtered result raises `AttributeError` rather than returning `{}`. `restrict_to_biallelic` / `restrict_to_segregating` route their empty case through `get_subset`, so they inherit this.

Two related sites, outside the strict variant-subset set:

- `GenotypeMatrix.from_haplotype_matrix` and `GenotypeMatrix.to_haplotype_matrix` drop both across the conversion, though the source carries them.
- `get_population_matrix` is a sample/row subset; it drops `samples` (which should be re-indexed to the selected samples) and `fields`.

#### Reference implementation

`HaplotypeMatrix.filter` and `GenotypeMatrix.filter` already do this correctly: they pass `samples=self.samples` and slice `fields` with a host-side index (`keep.get()`) in both the non-empty and empty-result branches. The fix is to bring the other methods to the same behaviour.

#### Proposed fix

Route the variant-subset family (`get_subset`, `get_subset_from_range`, `restrict_to_biallelic`, `restrict_to_segregating` on both classes) through a single shared subset helper that preserves `samples` and host-slices `fields`, mirroring `filter`. Set `.fields` in every empty-result branch so `.fields` access never raises. Decide the conversion (`from_haplotype_matrix` / `to_haplotype_matrix`) and `get_population_matrix` cases explicitly; note that a sample/row subset must also reshape any per-genotype `(n_var, n_samples)` field.

nspope:
Closes #181 

`get_subset`, `get_subset_from_range`, and the four methods sharing `_keep_sites` (`restrict_to_biallelic`, `restrict_to_segregating`, `exclude_missing_sites`, `filter_variants_by_missing`) built their result via a bare constructor call that dropped samples and fields entirely, unlike `filter()` which already carries both correctly. Also fixes `_empty_subset` to keep field keys with shape-0 arrays (matching `filter()`'s empty branch) instead of dropping them to an empty dict.

`restrict_to_segregating` dropped fields (samples was already carried). `from_haplotype_matrix` and `to_haplotype_matrix` dropped both samples and fields entirely, and `to_haplotype_matrix` also dropped `sample_sets` despite having everything needed to remap it (individual indices to haplotype pairs, the inverse of `from_haplotype_matrix`'s collapse). All three now match the existing correct behavior in `filter()`.

Followup issue: `get_population_matrix` requires re-indexing samples, and so is different in form (documented in #308)


pg_gpu/haplotype_matrix.py
```diff
@@ -1168,10 +1168,19 @@ def _empty_subset(self) -> "HaplotypeMatrix":
...
         return result
 
+    def _sliced_fields(self, keep_idx):
+        """self.fields sliced on axis 0 by keep_idx.
+
+        fields are host numpy arrays regardless of matrix device, so the
+        index is moved to host first.
+        """
+        keep_idx_np = keep_idx.get() if hasattr(keep_idx, 'get') else keep_idx
+        return {tag: arr[keep_idx_np] for tag, arr in self.fields.items()}
```
andrewkern, line 1182: This slices `self.fields` with indices that come from the masked view, but fields is on the underlying variant space, so the field values come out misaligned with positions. 

i think the right thing to do is to use `_accessible_idx` -- `filter()` does this correctly

pg_gpu/haplotype_matrix.py
```diff
@@ -1168,10 +1168,19 @@ def _empty_subset(self) -> "HaplotypeMatrix":
...
         result.accessible_mask = None
         return result
 
+    def _sliced_fields(self, keep_idx):
+        """self.fields sliced on axis 0 by keep_idx.
+
+        fields are host numpy arrays regardless of matrix device, so the
+        index is moved to host first.
+        """
+        keep_idx_np = keep_idx.get() if hasattr(keep_idx, 'get') else keep_idx
```
andrewkern, line 1181: `get` runs before we check if there are any fields. most matrices will have none. we could return {} early when self.fields is empty

pg_gpu/genotype_matrix.py
```diff
@@ -380,10 +380,16 @@ def from_haplotype_matrix(cls, hap_matrix):
                 ind_indices = sorted(set(i // 2 for i in indices))
                 new_sample_sets[name] = ind_indices
 
+        # fields are host numpy arrays; biallelic is on GPU.
+        new_fields = {tag: arr[biallelic.get()]
+                     for tag, arr in hap_matrix.fields.items()}
```
andrewkern, line 385: same issue as in haplotype matrix -- this will potentially be misaligned. biallelic is view length, but fields is underlying length. A masked HaplotypeMatrix with fields will crash.

pg_gpu/genotype_matrix.py
```diff
@@ -415,10 +421,21 @@ def to_haplotype_matrix(self):
...
+            }
+
         return HaplotypeMatrix(hap, self.positions, self.chrom_start,
                                self.chrom_end,
                                n_total_sites=self.n_total_sites,
-                               accessible_mask=self.accessible_mask)
+                               accessible_mask=self.accessible_mask,
+                               sample_sets=new_sample_sets,
+                               samples=self.samples,
+                               fields=self.fields)
```
andrewkern, line 438: `self.fields` goes in unfiltered. if a mask hides variants, the new HaplotypeMatrix has more field rows than variant columns. A later fi│lter() on it will return garbage or an error.

also -- this passes the arrays by reference. every other path copies

pg_gpu/genotype_matrix.py
```diff
@@ -898,10 +915,13 @@ def restrict_to_segregating(self):
         keep_idx = cp.where(keep)[0]
         new_geno = self.genotypes[:, keep_idx]
         new_pos = self.positions[keep_idx]
+        # fields are host numpy arrays; keep_idx is on GPU.
+        new_fields = {tag: arr[keep_idx.get()] for tag, arr in self.fields.items()}
```
andrewkern, line 919: this is the same class of bug as above. `keep_idx` is computed on the masked genotypes view

pg_gpu/haplotype_matrix.py
andrewkern, line 1698: we keep using this pattern, and it's buggy -- we should have a helper function here

andrewkern, line 1685: doesn't `_empty_subset()` do this code block?

tests/test_haplotype_matrix.py
```diff
@@ -309,6 +309,57 @@ def test_get_subset(sample_vcf):
...
+    return cp.asnumpy(arr) if isinstance(arr, cp.ndarray) else np.asarray(arr)
+
+
+def _assert_fields_match_keep(original, result):
+    """result's fields equal original's fields sliced to result's positions."""
+    orig_pos = _host(original.positions)
+    keep = np.searchsorted(orig_pos, _host(result.positions))
+    for tag, arr in original.fields.items():
+        np.testing.assert_array_equal(result.fields[tag], arr[keep])
+
```
andrewkern, line 337: we need a test that set an accessibility mask to reveal the bugs above -- maybe add that first, let it go red, then fix?

andrewkern (changes requested):
found a couple bugs here and have a few suggestions to simplify harden our code

nspope:
Thanks, good catch, I think I've addressed the misalignment risk now

andrewkern:
okay thanks @nspope. this looks ready. I've filed a follow up issue (#313) with a little additional clean up that I will handle now, but I'm going to merge this.

PR #309, opened 2026-09-15
Per-site decomposition for Weir-Cockerham FST

### LINKED ISSUE #187: Weir-Cockerham has no shared per-site decomposition, so the scatter engine cannot compute fst_wc

nspope:
Fixes #187 

fst_weir_cockerham already built per-site-per-allele variance components (a, b, c) before collapsing them to two floats inline, but nothing else could reuse that decomposition. Extracted _wc_site_components (mirroring diversity._ac_contribution's shared-decomposition pattern) returning the per-site (a, a+b+c) sums; the scalar function is now sum(a_site) / sum(abc_site).

The scatter engine (_windowed_twopop_scatter) gains an fst_wc branch built on the same function, plus the paired-rows warning check the fused kernel already had (its other statistics never needed one). fst_wc joins scatter_twopop, so windowed fst_wc now goes through the fast scatter path under both missing_data modes instead of only the fused kernel under 'include' and a slow per-window scalar loop under 'exclude'.


pg_gpu/windowed_analysis.py
```diff
@@ -1241,7 +1264,7 @@ def windowed_analysis(haplotype_matrix: HaplotypeMatrix,
     scatter_single = {'pi', 'theta_w', 'tajimas_d', 'segregating_sites',
                       'theta_h', 'theta_l', 'fay_wu_h', 'singletons',
                       'normalized_fay_wu_h', 'zeng_e', 'zeng_dh', 'max_daf'}
-    scatter_twopop = {'fst', 'fst_hudson', 'dxy', 'da'}
+    scatter_twopop = {'fst', 'fst_hudson', 'fst_wc', 'dxy', 'da'}
```
andrewkern, line 1267: so in looking at this closely, i think he scatter check runs before the fused check, so every `fst_wc` request under 'include' now goes through scatter, not the fused path. The fused path chunks when the matrix is big, and this won't. obviously preexisting, but maybe we should file an issue here? or fix it on this PR?

pg_gpu/windowed_analysis.py
```diff
@@ -1032,6 +1051,10 @@ def scatter_sum(values):
     if 'da' in stats_set:
         results['da'] = ((between_sum - (pi1_sum + pi2_sum) / 2.0) / spans_gpu).get()
 
+    if 'fst_wc' in stats_set:
+        results['fst_wc'] = cp.where(wc_abc_sum > 0, wc_a_sum / wc_abc_sum,
```
andrewkern, line 1055: Under 'exclude' this used to go through the scalar loop, which gives 0.0 when a+b+c is zero. Now it gives NaN. That matches the fused kernel and Hudson, so I think NaN is right, but it is a behavior change. prob worth adding a line to the changelog.

The scalar still returns 0.0 for the same case. So windowed and scalar now disagree on empty or monomorphic windows. Could be a follow-up, but worth an issue.

pg_gpu/windowed_analysis.py
```diff
@@ -936,7 +936,22 @@ def _windowed_twopop_scatter(haplotype_matrix, window_size, step_size,
     if haplotype_matrix.device == 'CPU':
         haplotype_matrix.transfer_to_gpu()
 
+    stats_set = set(statistics)
     pop1_name, pop2_name = populations[0], populations[1]
+
+    if 'fst_wc' in stats_set:
```
andrewkern, line 942: This block is the same as the one in `windowed_statistics_fused`, and the chunked engine has a third. One small helper in `_warnings.py` that takes the matrix and the two pops would replace all three.

pg_gpu/windowed_analysis.py
```diff
@@ -1002,7 +1017,11 @@ def scatter_sum(values):
     # Compute per-site components (single pass over the data)
     mpd1, mpd2, between = _twopop_site_components(hap1, hap2)
```
andrewkern, line 1018: This runs even when the request is only `fst_wc`. we should limit it to the stats that use it. And `_wc_site_components` recomputes `k` with two more `.max()` calls. Pass `k` in.

tests/test_windowed_analysis.py
```diff
@@ -1307,6 +1367,11 @@ def test_scatter_dxy_fst_da_match_scalar(self, sim_hm):
             assert np.isclose(da_w,
                               divergence.da(sub, 'p1', 'p2', span_normalize=False),
                               rtol=1e-9, atol=1e-12)
+            wc_ref = divergence.fst_weir_cockerham(sub, 'p1', 'p2')
+            if np.isnan(wc_w) or np.isnan(wc_ref):
+                assert np.isnan(wc_w) and np.isnan(wc_ref)
```
andrewkern, line 1372: The "both NaN" branch can never pass. `fst_weir_cockerham` never returns NaN. So if a window ever comes out NaN this fails with a confusing message. Either make the scalar return NaN too or drop the branch. Same at line 693.

tests/test_windowed_analysis.py
```diff
@@ -497,6 +497,30 @@ def test_two_pop_chunked_matches_fused(self, matrix_with_pops):
                                        equal_nan=True,
                                        err_msg=f"Mismatch in {k}")
 
+    def test_two_pop_scatter_matches_fused_for_fst_wc(self, matrix_with_pops):
+        """The scatter engine (now handling fst_wc) must agree exactly with
```
andrewkern, line 501: "now handling fst_wc" is history. Say what the test checks, not what changed. Same for "now that fst_wc is scatter-eligible" at line 663.

pg_gpu/divergence.py
```diff
@@ -372,10 +335,8 @@ def fst_weir_cockerham(haplotype_matrix,
     p_bar[vt] = (n1t * p1[vt] + n2t * p2[vt]) / nt[:, None]
     s_squared[vt] = (n1t * (p1[vt] - p_bar[vt])**2
                      + n2t * (p2[vt] - p_bar[vt])**2) / ((r - 1) * (nt / r)[:, None])
-    h_bar[vt] = (het1[vt] + het2[vt]) / nt[:, None]       # per-allele obs het
+    h_bar[vt] = (het1[vt] + het2[vt]) / nt[:, None]
```
andrewkern, line 338: this dropped the comments that link the code to the paper. better to keep them in i reckon

andrewkern (changes requested):

andrewkern:
small nit picks here.

nspope:
So I think there's some overlap here with #311 --- it fixes the NaN issues you raise. I'll deal with the other issues now, I think.

pg_gpu/windowed_analysis.py
nspope: Let's defer this as it's pre-existing, i'll file a followup shortly.

nspope:
OK, I think is good, with deferrals mentioned above. Maybe take a look at #311 next given the overlap?

andrewkern:
looks good. i'm pushing one small change to close out all of 187, then will merge and move to 311.

PR #311, opened 2026-09-18
Make undefined Fst NaN, and fix multi-population single-stat column names

### LINKED ISSUE #200: Engines disagree on the undefined-FST sentinel and on single-population columns in mixed requests

nspope:
Closes #200 

`fst_hudson`/`fst_weir_cockerham` returned 0.0 for an undefined ratio (no site with data in both populations, or no polymorphism at all) while every windowed engine's own Fst math already used `nan` for the same case; the per-window Python-loop fallback inherited the scalar 0.0, so it silently disagreed with the fused/scatter engines for the identical window. Standardize on `nan`, matching `divergence.gmin`'s existing convention.

Separately, `windowed_analysis()` collapsed a single-population statistic (e.g. pi) onto one bare column whenever 2+ populations were named alongside a two-population statistic, silently keeping only the first population's value -- both under 'include' (via the fused engine's `population=pop1`) and under 'exclude' for `fst`/`fst_hudson`/`dxy`/`da` (via a scatter dispatch branch that also only read the first population). Route any single-pop statistic requested with 2+ named populations through a new per-population loop that suffixes each column "<stat>_<pop>", matching the naming already used elsewhere, and recurse for whatever else was requested.


pg_gpu/windowed_analysis.py
```diff
@@ -1245,46 +1274,67 @@ def windowed_analysis(haplotype_matrix: HaplotypeMatrix,
...
+                    statistics=rest,
+                    populations=populations,
+                    missing_data=missing_data,
+                    span_normalize=span_normalize,
+                    accessible_bed=None,  # already applied above
+                    chrom=chrom,
+                    **kwargs,
+                )
+                if df2 is not None and not df2.empty:
+                    for col in df2.columns:
```
andrewkern: This assumes df2 has the same rows as df1. It does not when the recursion lands in the Python-loop fallback, which only emits windows that contain a variant, while the scatter engine emits every grid window. With 3 populations, `['pi', 'fst']`, and one empty window this raises `ValueError: Length of values (4) does not match length of index (6)`. That call worked before this PR. Reproduced.

pg_gpu/windowed_analysis.py
```diff
@@ -1245,46 +1274,67 @@ def windowed_analysis(haplotype_matrix: HaplotypeMatrix,
...
+        # else was requested recurses back through this function; the
+        # remainder excludes every scatter_single member by construction, so
+        # it can never re-enter this branch.
+        single_part = sorted(requested & scatter_single)
+        if single_part and n_pops >= 2:
+            df1 = _per_population_scatter_singles(
+                haplotype_matrix, window_size, step_size,
+                single_part, populations, missing_data, span_normalize,
+                chrom=chrom)
+            rest = sorted(requested - set(single_part))
```
andrewkern: `sorted` here breaks on a callable statistic. `['pi', 'fst', my_fn]` with two populations raises `TypeError: '<' not supported between instances of 'str' and 'function'`. Before this PR that request fell through to the fallback, which accepts callables. Reproduced.

pg_gpu/windowed_analysis.py
```diff
@@ -1245,46 +1274,67 @@ def windowed_analysis(haplotype_matrix: HaplotypeMatrix,
...
+        # routing to the fused engine with just the first population, and
+        # 'exclude' does the same whenever the other requested statistic is
+        # one _windowed_twopop_scatter also handles (fst/fst_hudson/dxy/da) --
+        # only a twopop stat outside that set (fst_wc) happens to fall
+        # through to the correct per-population fallback on its own. Whatever
+        # else was requested recurses back through this function; the
+        # remainder excludes every scatter_single member by construction, so
+        # it can never re-enter this branch.
+        single_part = sorted(requested & scatter_single)
+        if single_part and n_pops >= 2:
```
andrewkern: This only covers `scatter_single`. `garud_h1` or `mean_nsl` with two populations still comes back as one bare column holding the first population's value, no warning. Issue 200 asks for one rule in every engine. Right now the quickstart's everything-at-once example returns `pi_pop1`, `pi_pop2` next to a bare `garud_h1`.

pg_gpu/windowed_analysis.py
```diff
@@ -880,6 +880,35 @@ def windowed_test(w1, w2, numerator_arr):
...
+    fallback already uses whenever 2+ populations are named. Callers with 0
+    or 1 population should call ``_windowed_thetas_scatter`` directly instead
+    -- its bare-name convention is unambiguous there.
+    """
+    merged = None
+    for pop in populations:
+        df = _windowed_thetas_scatter(
+            haplotype_matrix, window_size, step_size,
+            statistics, [pop], missing_data, span_normalize, chrom=chrom)
+        if df is None:
```
andrewkern: `_windowed_thetas_scatter` never returns None. It returns an empty frame when a population has no complete sites under 'exclude'. Then this loop just skips that population, so you get `pi_p1` and no `pi_p2` with no error. Reproduced. Either give the missing population a NaN column or raise.

pg_gpu/windowed_analysis.py
```diff
@@ -880,6 +880,35 @@ def windowed_test(w1, w2, numerator_arr):
...
+            return None
+        df = df.rename(columns={
+            stat: f"{stat}_{pop}" for stat in statistics if stat in df.columns
+        })
+        if merged is None:
+            merged = df
```
andrewkern: The merged frame keeps only the first population's `n_variants`. Under 'exclude' each population is filtered to its own site set, so that count is wrong for the other columns. Also, this copy-columns loop now exists three times in this file. One small helper that checks row counts would cover all of them and would have caught the crash above.

pg_gpu/divergence.py
```diff
@@ -177,7 +177,7 @@ def fst_hudson(haplotype_matrix: HaplotypeMatrix,
     valid_mask = den > 0
     if cp.any(valid_mask):
         return float((cp.sum(num[valid_mask]) / cp.sum(den[valid_mask])).get())
-    return 0.0
+    return float('nan')
```
andrewkern: okay so now `fst(method='hudson')` disagrees with `fst_nei` and `fst_tskit` which still return 0.0 for this case - can we add those two as well to this PR?

tests/test_implementation_parity.py
```diff
@@ -495,3 +496,54 @@ def test_daf_hist_matches_scalar(request, condition):
...
+                           missing_data=missing_data)
+    pi2_ref = diversity.pi(hm, population=POP2, span_normalize=False,
+                           missing_data=missing_data)
+    fst_wc_ref = divergence.fst_weir_cockerham(hm, POP1, POP2,
+                                               missing_data=missing_data)
+
+    assert np.isclose(df["pi_pop1"].iloc[0], pi1_ref, rtol=RTOL, atol=ATOL)
+    assert np.isclose(df["pi_pop2"].iloc[0], pi2_ref, rtol=RTOL, atol=ATOL)
+    assert np.isclose(df["fst_wc"].iloc[0], fst_wc_ref, rtol=RTOL, atol=ATOL)
+    # The regression this guards against silently dropped pop2 and returned
```
andrewkern: This comment describes the old bug. Say what the check is for instead: pi must differ between the two populations or a collapsed column would go unnoticed.

andrewkern (changes requested):
caught some bugs here.

pg_gpu/windowed_analysis.py
nspope: in #310 I made the site set shared across populations (same site set for dxy, pi1, pi2). But perhaps that's the wrong choice?

nspope:
Looking into this: the issues here largely stem from trying to merge the python-loop fallback (skips windows that are empty) with the scatter/fused engines (emit empty windows with nan or 0). That's different numbers of rows with empty windows, so problems ensue. Whether there should actually be this asymmetry in behavior is a good question, and I don't think I'll try to fix it here. I'll see if I can salvage this PR for the original NaN-vs-zero fix, if not will close and try to break up into multiple issues.

pg_gpu/windowed_analysis.py
andrewkern: ah okay! maybe i should have reviewed #310 first. using the same site set seems a reasonable choice to me. sorry about that

nspope: not sure order matters (there are likely bugs there too!) but it's probably easiest to keep this PR scoped to just the NaNs for now, and think more carefully about the other half of the issue (in another PR)

